PR #4, opened 2023-11-17
Go-to setpoint type

tstastny:
Adds a "Go-to" setpoint type which streams a position and (optional) heading setpoint, meant to be handled by a position and heading smoother on the flight controller. Main idea: dump in whatever position and heading you want, and they will be smoothly followed by the autopilot. Additional fields for velocity and heading rate limits are included.

An example mode using the Go-to setpoint type is included:
https://github.com/Auterion/px4-ros2-interface-lib/assets/8026163/76795ebc-13b9-4637-b7f7-867e773047fa

Depends on: https://github.com/PX4/PX4-Autopilot/pull/22375

Docs: https://github.com/PX4/PX4-user_guide/pull/2935

### DIFF
--- a/examples/cpp/modes/goto/include/mode.hpp
+++ b/examples/cpp/modes/goto/include/mode.hpp
@@ -1,0 +1,222 @@
+/****************************************************************************
+ * Copyright (c) 2023 PX4 Development Team.
+ * SPDX-License-Identifier: BSD-3-Clause
+ ****************************************************************************/
+#pragma once
+
+#include "util.hpp"
+
+#include <px4_ros2/components/mode.hpp>
+#include <px4_ros2/control/setpoint_types/goto.hpp>
+#include <px4_ros2/odometry/local_position.hpp>
+#include <px4_msgs/msg/vehicle_attitude.hpp>
+
+#include <rclcpp/rclcpp.hpp>
+#include <Eigen/Core>
+#include <algorithm>
+
+static const std::string kName = "Go-to Example";
+static const std::string kNodeName = "example_mode_goto";
+
+class FlightModeTest : public px4_ros2::ModeBase
+{
+public:
+  explicit FlightModeTest(rclcpp::Node & node)
+  : ModeBase(node, kName)
+  {
+    _goto_setpoint = std::make_shared<px4_ros2::GotoSetpointType>(*this);
+
+    _vehicle_local_position = std::make_shared<px4_ros2::OdometryLocalPosition>(*this);
+
+    _vehicle_attitude_sub = node.create_subscription<px4_msgs::msg::VehicleAttitude>(
+      "/fmu/out/vehicle_attitude", rclcpp::QoS(1).best_effort(),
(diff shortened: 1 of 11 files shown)


examples/cpp/modes/goto/include/mode.hpp
```diff
@@ -0,0 +1,222 @@
...
+{
+public:
+  explicit FlightModeTest(rclcpp::Node & node)
+  : ModeBase(node, kName)
+  {
+    _goto_setpoint = std::make_shared<px4_ros2::GotoSetpointType>(*this);
+
+    _vehicle_local_position = std::make_shared<px4_ros2::OdometryLocalPosition>(*this);
+
+    _vehicle_attitude_sub = node.create_subscription<px4_msgs::msg::VehicleAttitude>(
```
bkueng, line 31: We should add this to the odometry

examples/cpp/modes/goto/include/util.hpp
```diff
@@ -0,0 +1,43 @@
+/****************************************************************************
+ * Copyright (c) 2023 PX4 Development Team.
+ * SPDX-License-Identifier: BSD-3-Clause
+ ****************************************************************************/
+#pragma once
+
+#include <Eigen/Eigen>
+#include <cmath>
+
+static inline Eigen::Vector3f quaternionToEuler(const Eigen::Quaternionf & q)
```
bkueng, line 10: These should be made generally available under some `math/conversion` dir

examples/cpp/modes/goto/include/mode.hpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+private:
+  static constexpr float kTriangleHeight = 20.F; // [m]
+  static constexpr float kTriangleWidth = 30.F; // [m]
+
+  enum class State
+  {
+    kSettlingAtStart = 0,
```
bkueng, line 150: enum's don't use the constants naming
```suggestion
    SettlingAtStart = 0,
```

bkueng (commented):
Looks good, we can also make the attitude & conversions generally available later on.

examples/cpp/modes/goto/include/mode.hpp
tstastny: done

examples/cpp/modes/goto/include/util.hpp
tstastny: yeah i agree .. just didnt do it yet for the sake of getting the example out quick - added ugly note to shame myself until we consider where to better move these utils

examples/cpp/modes/goto/include/mode.hpp
tstastny: agree - perhaps a follow up pr..

tstastny:
@bkueng made quick minimal changes - but the rest i think we just do in a follow up to enable folks quickly on a working example for goto, ok for you?

PR #5, opened 2023-11-20
External navigation interface

GuillaumeLaine:
Contents:
- Local navigation interface
- Example usage node of local navigation interface
- Global navigation interface
- Example usage node of global navigation interface

### Testing
- Flight tested by passing lat lon measurements to the global interface from `vehicle_gps_position` plus noise, and checking aux global position was fused in the EKF.
- Flight tested by passing 3D local position measurements to the local interface from `vehicle_local_position` plus noise, and checking external vision position was fused in the EKF.

### Usage
Build, source, run local interface example node which uses the interface to send a sample estimate every 1s:
```
ros2 run example_local_navigation_cpp example_local_navigation_cpp
```
Similarly for global interface example node:
```
 ros2 run example_global_navigation_cpp example_global_navigation_cpp
```


px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  _aux_local_position_pub = node.create_publisher<AuxLocalPosition>(AUX_LOCAL_POSITION_TOPIC, 10);
+
+}
+
+int LocalNavigationInterface::update(const LocalPositionEstimate & local_position_estimate) const
+{
+  // Run basic sanity checks on local position estimate
+  if (_checkEstimateEmpty(local_position_estimate)) {
+    RCLCPP_WARN(_node.get_logger(), "Estimate values are all empty.");
```
DanMesh, line 49: I wonder whether `WARN` is the right level for this. While it is bad, if a user sends an invalid struct, there's a good chance they're continually sending it at a high rate, which would spam the logs. That might be a good thing too, but we could also make this log `DEBUG` level (or even remove it) and rely more on the return value for error catching. What do you think?

px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
```diff
@@ -0,0 +1,201 @@
...
+  AuxLocalPosition aux_local_position;
+
+  aux_local_position.timestamp_sample = local_position_estimate.timestamp_sample;
+
+  // Position
+  aux_local_position.pose_frame = _pose_frame;
+
+  const Vector2f position_xy = local_position_estimate.position_xy.value_or(Vector2f{NAN, NAN});
+  aux_local_position.position[0] = position_xy[0];
+  aux_local_position.position[1] = position_xy[1];
```
DanMesh, line 75: Do we need some kind of check that both the `x` and `y` components have values? This goes for velocity, variances etc.
@bresch I would assume we want to avoid sending in partial positions, e.g. `position = [12.3, NAN]`.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_navigation_interface.hpp
```diff
@@ -0,0 +1,94 @@
...
+    AuxLocalPosition::POSE_FRAME_NED,
+    AuxLocalPosition::POSE_FRAME_FRD
+  };
+
+  uint8_t _available_velocity_frames[4] = {
+    AuxLocalPosition::VELOCITY_FRAME_UNKNOWN,
+    AuxLocalPosition::VELOCITY_FRAME_NED,
+    AuxLocalPosition::VELOCITY_FRAME_FRD,
+    AuxLocalPosition::VELOCITY_FRAME_BODY_FRD
+  };
```
DanMesh, line 90: Could these be const and/or static?

DanMesh (commented):
This is off to a good start. I looked mostly at the local updates for now and left a few thoughts.

px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
```diff
@@ -0,0 +1,201 @@
...
+    local_position_estimate.velocity_z_variance.value_or(NAN);
+
+  // Angular velocity (unused at the moment)
+  aux_local_position.angular_velocity = {NAN, NAN, NAN};
+
+  // Publish
+  aux_local_position.timestamp = _node.get_clock()->now().nanoseconds() * 1e-3;
+  _aux_local_position_pub->publish(aux_local_position);
+
+  return static_cast<int>(NavigationInterfaceCodes::SUCCESS);
```
GuillaumeLaine, line 121: @DanMesh just realized, maybe we should return a NavigationInterfaceCodes directly? It would prevent type casting, not sure if there's any advantage to returning an int?

px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
GuillaumeLaine: I see your point. I'll set it to DEBUG in the meantime.

px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
GuillaumeLaine: That's a good point, I'll add a check for this.

.gitignore
```diff
@@ -380,3 +380,11 @@ pyrightconfig.json
...
 # End of https://www.toptal.com/developers/gitignore/api/c++,meson,sonar,python,sonarqube,visualstudiocode,clion
 .idea
+
+# Colcon
+build/
+install/
+log/
+
+# Vscode
+.vscode
```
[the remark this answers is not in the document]
GuillaumeLaine: You're correct removing it

px4_ros2_cpp/include/px4_ros2/navigation/experimental/global_navigation_interface.hpp
```diff
@@ -0,0 +1,77 @@
...
+  explicit GlobalNavigationInterface(Context & context);
+  ~GlobalNavigationInterface() override = default;
+
+  /**
+   * @brief Publish global position estimate to FMU.
+   */
+  NavigationInterfaceReturnCode update(const GlobalPositionEstimate & global_position_estimate)
+  const override;
+
+  const std::string AUX_GLOBAL_POSITION_TOPIC = "/fmu/in/aux_global_position";
```
DanMesh, line 48: Remove?

px4_ros2_cpp/include/px4_ros2/navigation/experimental/navigation_interface_base.hpp
```diff
@@ -0,0 +1,34 @@
...
+   * @brief Publish position estimate to FMU.
+   */
+  virtual NavigationInterfaceReturnCode update(const EstimateType & global_position_estimate) const
+  = 0;
+
+private:
+  virtual bool _checkEstimateEmpty(const EstimateType & estimate) const = 0;
+  virtual bool _checkVarianceValid(const EstimateType & estimate) const = 0;
+  virtual bool _checkFrameValid(const EstimateType & estimate) const = 0;
+  virtual bool _checkValuesNotNAN(const EstimateType & estimate) const = 0;
```
DanMesh, line 31: On naming, how about something like `_isX`?

Could we also make it such that they all return the same value when "good"? For example, when "good", we currently have `_checkEstimateEmpty` = `false` while all the others are `true`. Maybe this isn't too important if they're not used together.

Open to ideas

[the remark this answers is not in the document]
GuillaumeLaine: Protected certainly makes more sense here

px4_ros2_cpp/include/px4_ros2/navigation/experimental/global_navigation_interface.hpp
```diff
@@ -0,0 +1,77 @@
...
+   * @brief Check that if an estimate value is defined, its associated frame is not *FRAME_UNKNOWN.
+   */
+  bool _checkFrameValid(const GlobalPositionEstimate & estimate) const override;
+
+  /**
+   * @brief Check that if an estimate value is defined, none of its fields are NAN.
+   */
+  bool _checkValuesNotNAN(const GlobalPositionEstimate & estimate) const override;
+
+  rclcpp::Node & _node;
```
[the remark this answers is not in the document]
GuillaumeLaine: Shouldn't this stay `private`? Given that we are not expecting children to the global navigation interface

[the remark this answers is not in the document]
DanMesh: My main thought here is that it should be in the interface base, since it is common to both global and local.
Then, if it's in the base, it would need to be protected so that the global & local children can access it.

examples/cpp/navigation/global_navigation/include/global_navigation.hpp
```diff
@@ -0,0 +1,96 @@
...
+using namespace std::chrono_literals; // NOLINT
+
+class GlobalNavigationTest : public px4_ros2::Context
+{
+public:
+  explicit GlobalNavigationTest(rclcpp::Node & node)
+  : Context(node), _node(node)
+  {
+    // Instantiate global navigation interface
+    _global_navigation_interface = std::make_shared<px4_ros2::GlobalNavigationInterface>(*this);
```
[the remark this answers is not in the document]
GuillaumeLaine: I agree that for the user just passing a `Node` to the interface to get things running is ideal. But In that case if we're not passing a `Context` as a constructor argument to the interface, how will the interface get hold of a Context? I believe one of the future uses of the Context for the interface had to do with registration with the FMU. But maybe this can be done directly in the interface, though I think that would be inconsistent with the modes implementation.

@bkueng maybe you have some insight on this? Does it make sense to have the interface just take a `Node` for initialisation?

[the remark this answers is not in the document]
GuillaumeLaine: Or perhaps the interface can inherit from `Context` in that case? Then the Context is abstracted away, but still available internally

px4_ros2_cpp/include/px4_ros2/navigation/experimental/global_navigation_interface.hpp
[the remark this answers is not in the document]
GuillaumeLaine: Ah duh, I didn't see the word base, will do

examples/cpp/navigation/global_navigation/include/global_navigation.hpp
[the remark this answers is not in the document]
bkueng: Yes the idea is to do something like this:
Library:
```
class NavigationInferaceBase : public Context {
...

class GlobalNavigationInferaceBase : public NavigationInferaceBase {
...
```
User:
```
class GlobalNavigationTest : public px4_ros2::GlobalNavigationInferaceBase {
...
```

examples/cpp/navigation/global_navigation/include/global_navigation.hpp
```diff
@@ -0,0 +1,68 @@
...
+    px4_ros2::GlobalPositionEstimate global_position_estimate {};
+
+    global_position_estimate.timestamp_sample = _node.get_clock()->now();
+
+    global_position_estimate.lat_lon = Eigen::Vector2f {12.34321, 23.45432};
+    global_position_estimate.altitude_agl = 12.4f;
+    global_position_estimate.positional_uncertainty = 0.4f;
+
+    px4_ros2::NavigationInterfaceReturnCode retcode;
+    retcode = update(global_position_estimate);
```
DanMesh, line 39: Is it necessary to split these?

px4_ros2_cpp/include/px4_ros2/navigation/experimental/global_navigation_interface.hpp
```diff
@@ -0,0 +1,76 @@
...
+{
+
+struct GlobalPositionEstimate
+{
+  rclcpp::Time timestamp_sample {};
+
+  // Lat lon
+  std::optional<Vector2d> lat_lon {std::nullopt};
+  // Standard deviation of horizontal position error (metres)
+  std::optional<float> lat_lon_stddev {std::nullopt};
```
GuillaumeLaine, line 28: @DanMesh Is this the name we want to expose `lat_lon_stddev`? Or should we use `eph` instead. The former is more descriptive, but the latter is closer to the VehicleGlobalPosition message.

GuillaumeLaine: @bresch any opinion on this?

bresch: I think eph and epv are widely used

DanMesh: What about using a variance instead of a standard deviation? e.g. `lat_lon_variance` and `altitude_variance`.

This would then be consistent with the local navigation interface, which uses variances.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/global_navigation_interface.hpp
```diff
@@ -0,0 +1,76 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ ****************************************************************************/
+
+#pragma once
+
+#include <optional>
+#include <rclcpp/rclcpp.hpp>
+#include <Eigen/Eigen>
+
+#include <px4_msgs/msg/vehicle_global_position.hpp>
```
DanMesh, line 12: We should add the topics used to the message compatibility checks. See https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/components/message_compatibility_check.hpp#L12-L33.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/global_navigation_interface.hpp
GuillaumeLaine: discussion decision: use `horizontal_variance` and `vertical_variance`

px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
```diff
@@ -0,0 +1,281 @@
...
+    std::begin(kAvailableVelocityFrames), std::end(
+      kAvailableVelocityFrames), velocity_frame);
+  if (it_vel_frame == std::end(kAvailableVelocityFrames)) {
+    RCLCPP_WARN(
+      node.get_logger(), "Failed to initialize LocalNavigationInterface: invalid velocity reference frame %i. Setting velocity reference frame to VELOCITY_FRAME_UNKNOWN.",
+      velocity_frame);
+    _velocity_frame = AuxLocalPosition::VELOCITY_FRAME_UNKNOWN;
+  } else {
+    _velocity_frame = velocity_frame;
+  }
```
bkueng, line 40: Defining a separate enum would make these a bit nicer to use (instead of uint8_t), with the small downside of having to map from the enum to `AuxLocalPosition::POSE_FRAME_x`.

bkueng (commented):
Nice work, I remarked a couple of smaller things.

.gitignore
```diff
@@ -380,3 +380,8 @@ pyrightconfig.json
 
 # End of https://www.toptal.com/developers/gitignore/api/c++,meson,sonar,python,sonarqube,visualstudiocode,clion
 .idea
+
+# Colcon
```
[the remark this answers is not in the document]
GuillaumeLaine: Because I was calling `colcon build` directly from the interface lib directory. But I guess the entire lib should be in a parent ros workspace, and I should call `colcon build` from there? I'll remove the changes from the .gitignore

examples/cpp/navigation/local_navigation/include/local_navigation.hpp
```diff
@@ -0,0 +1,76 @@
...
+    const uint8_t velocity_frame)
+  : LocalNavigationInterface(node, pose_frame, velocity_frame)
+  {
+    _timer =
+      node.create_wall_timer(1s, [this] {updateAuxLocalPosition();});
+
+    RCLCPP_INFO(node.get_logger(), "example_local_navigation_node running!");
+  }
+
+  void updateAuxLocalPosition()
```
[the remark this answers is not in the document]
GuillaumeLaine: It does not, i'll update accordingly

px4_ros2_cpp/include/px4_ros2/navigation/experimental/navigation_interface_base.hpp
```diff
@@ -0,0 +1,68 @@
...
+{
+public:
+  explicit NavigationInterfaceBase(rclcpp::Node & node)
+  : Context(node, ""), _node(node) {}
+  virtual ~NavigationInterfaceBase() = default;
+
+  /**
+   * @brief Publish position estimate to FMU.
+   */
+  virtual NavigationInterfaceReturnCode update(const EstimateType & global_position_estimate) const
```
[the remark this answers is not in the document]
GuillaumeLaine: I thought it made sense because the `update` method is the core shared method between all the children navigation interfaces. Removing the template would also mean removing all the templated bool check methods from the base class (e.g. `isEstimateNonEmpty`...), and the base case would then only be a `Context` constructor call. Do you think it's best to remove the update method?

px4_ros2_cpp/src/navigation/experimental/global_navigation_interface.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+  _aux_global_position_pub =
+    node.create_publisher<VehicleGlobalPosition>(
+    topicNamespacePrefix() + "/fmu/in/aux_global_position", 10);
+}
+
+NavigationInterfaceReturnCode GlobalNavigationInterface::update(
+  const GlobalPositionEstimate & global_position_estimate) const
+{
+  // Run basic sanity checks on global position estimate
```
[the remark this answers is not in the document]
GuillaumeLaine: Do you mean in the PX4 user guide? Each check has a `@brief` description, but how do you mean otherwise?

examples/cpp/navigation/global_navigation/include/global_navigation.hpp
```diff
@@ -0,0 +1,69 @@
...
+    // Enable debug output
+    auto ret =
+      rcutils_logging_set_logger_level(get_logger().get_name(), RCUTILS_LOG_SEVERITY_DEBUG);
+
+    if (ret != RCUTILS_RET_OK) {
+      RCLCPP_ERROR(get_logger(), "Error setting severity: %s", rcutils_get_error_string().str);
+      rcutils_reset_error();
+    }
+
+    _interface = std::make_unique<GlobalNavigationTest>(*this);
```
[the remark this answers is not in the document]
DanMesh: I guess that would go in the interface base class, rather than each of the global and local child classes?

px4_ros2_cpp/src/navigation/experimental/global_navigation_interface.cpp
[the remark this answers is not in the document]
DanMesh: > Ah or do you mean to extend the @brief description of the update method?

I'd do this, to have it in the code. But mentioning it in the user guide is also a good idea

examples/cpp/navigation/global_navigation/include/global_navigation.hpp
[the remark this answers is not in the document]
bkueng: Exactly

px4_ros2_cpp/src/navigation/experimental/global_navigation_interface.cpp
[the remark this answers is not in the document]
bkueng: Exactly, for the `update`, since that is what someone using the API most likely will look at (you could also link to individual checks if there's more text to avoid repetition)

px4_ros2_cpp/src/navigation/experimental/local_navigation_interface.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+bool LocalNavigationInterface::isFrameValid(const LocalPositionEstimate & estimate) const
+{
+  if ((estimate.position_xy.has_value() || estimate.position_z.has_value()) &&
+    _pose_frame == AuxLocalPosition::POSE_FRAME_UNKNOWN)
+  {
+    RCLCPP_ERROR(
+      _node.get_logger(),
+      "Position estimate update requested but pose reference frame is set to POSE_FRAME_UNKNOWN.");
```
DanMesh, line 170: I would try abbreviate this. Also, this function can in theory be called when you're not requesting an update.
Same applies to velocity below, and the error string in the invalid argument exception ([here](https://github.com/Auterion/px4-ros2-interface-lib/pull/5/files#diff-19dae766cc08d36ce42531a0f6f4460ad4736ada1694ddfd54eb1c9d4e04deccR40)).

```suggestion
      "Position estimate has unknown pose frame.");
```

DanMesh (approved):
The [aux global position PR](https://github.com/PX4/PX4-Autopilot/pull/22426) is merged, so this should be ready to go

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_navigation_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+  /** @brief Variance of velocity error in the xy-plane. */
+  std::optional<Eigen::Vector2f> velocity_xy_variance {std::nullopt};
+  /** @brief Velocity in the z-axis. */
+  std::optional<float> velocity_z {std::nullopt};
+  /** @brief Variance of velocity error in the z-axis. */
+  std::optional<float> velocity_z_variance {std::nullopt};
+
+  /** @brief Attitude quaternion. */
+  std::optional<Eigen::Quaternionf> attitude_quaternion {std::nullopt};
+  /** @brief Variance of attitude error. */
```
bresch, line 95:
```suggestion
  /** @brief Variance of attitude in body frame. */
```
It's also possible to define the attitude variance in the local frame, so we should make it explicit.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_navigation_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+  /** @brief Velocity in the xy-plane. */
+  std::optional<Eigen::Vector2f> velocity_xy {std::nullopt};
+  /** @brief Variance of velocity error in the xy-plane. */
+  std::optional<Eigen::Vector2f> velocity_xy_variance {std::nullopt};
+  /** @brief Velocity in the z-axis. */
+  std::optional<float> velocity_z {std::nullopt};
+  /** @brief Variance of velocity error in the z-axis. */
+  std::optional<float> velocity_z_variance {std::nullopt};
+
+  /** @brief Attitude quaternion. */
```
bresch, line 93:
```suggestion
  /** @brief Attitude quaternion [w, x, y, z], Hamiltonian convention */
```

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_navigation_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+    case VelocityFrame::FRD: return AuxLocalPosition::VELOCITY_FRAME_FRD;
+
+    case VelocityFrame::BodyFRD: return AuxLocalPosition::VELOCITY_FRAME_BODY_FRD;
+  }
+
+  return AuxLocalPosition::POSE_FRAME_UNKNOWN;
+}
+
+/**
+ * @struct LocalPositionEstimate
```
bresch, line 64:
```suggestion
 * @struct LocalPositionMeasurement
```

Nitpicking here, but using the name "Estimate" is a bit confusing here as we should be sending "measurements" to the estimator and getting back "estimates". Another possible shorter name is `LocalPositionSample`.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_navigation_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+}
+
+constexpr inline uint8_t velocityFrameToMessageFrame(px4_ros2::VelocityFrame frame) noexcept
+{
+  switch (frame) {
+    case VelocityFrame::Unknown: return AuxLocalPosition::VELOCITY_FRAME_UNKNOWN;
+
+    case VelocityFrame::NED: return AuxLocalPosition::VELOCITY_FRAME_NED;
+
+    case VelocityFrame::FRD: return AuxLocalPosition::VELOCITY_FRAME_FRD;
```
bresch, line 55:
```suggestion
    case VelocityFrame::LocalFRD: return AuxLocalPosition::VELOCITY_FRAME_FRD;
```
Most people are using "FRD" for "body frame". I know that you just took the uorb message as a reference, but I think we should be a bit more explicit.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_navigation_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+  return AuxLocalPosition::POSE_FRAME_UNKNOWN;
+}
+
+constexpr inline uint8_t velocityFrameToMessageFrame(px4_ros2::VelocityFrame frame) noexcept
+{
+  switch (frame) {
+    case VelocityFrame::Unknown: return AuxLocalPosition::VELOCITY_FRAME_UNKNOWN;
+
+    case VelocityFrame::NED: return AuxLocalPosition::VELOCITY_FRAME_NED;
```
bresch, line 53:
```suggestion
    case VelocityFrame::LocalNED: return AuxLocalPosition::VELOCITY_FRAME_NED;
```
Yes, NED is necessary a local frame by definition, but it's then more obvious that the user has the choice between 2 local frames and a body frame

bresch (changes requested):

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_position_measurement_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+  FRD
+};
+
+enum class VelocityFrame
+{
+  Unknown,
+  LocalNED,
+  LocalFRD,
+  BodyFRD
+};
```
DanMesh, line 33: @bresch Do you think we should also make the _pose_ frames `LocalNED` and `LocalFRD`? While there is no chance for confusion between local and body, it would be more consistent with the velocity frames.

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_position_measurement_interface.hpp
```diff
@@ -0,0 +1,145 @@
...
+   */
+  bool isVarianceValid(const LocalPositionMeasurement & measurement) const;
+
+  /**
+   * @brief Check that if an measurement value is defined, its associated frame is not *FRAME_UNKNOWN.
+   */
+  bool isFrameValid(const LocalPositionMeasurement & measurement) const;
+
+  /**
+   * @brief Check that if an measurement value is defined, none of its fields are NAN.
```
DanMesh, line 135:
```suggestion
   * @brief Check that if a measurement value is defined, none of its fields are NAN.
```

Being a bit pedantic, but find-and-replace unfortunately leaves behind the old articles for every instance of "measurement"

DanMesh (changes requested):

px4_ros2_cpp/include/px4_ros2/navigation/experimental/local_position_measurement_interface.hpp
bresch: Yes, makes total sense

DanMesh:
Great, thanks @bkueng!

PR #7, opened 2023-11-28
Doxygen configuration & deploy to github pages

bkueng:
@hamishwillee I was playing a bit with doxygen and we can add something automated with minimal effort that I think can be useful.
This is how it looks like: https://auterion.github.io/px4-ros2-interface-lib/modules.html

Let me know what you think.


hamishwillee:
I think it is useful as is. A few comments for easy-wins.

1. A cross link to the PX4 docs.
2. The modules section lists https://auterion.github.io/px4-ros2-interface-lib/group__Setpoints.html under controls. Other experimental setpoint types are not listed here, but are in the class list - such as https://auterion.github.io/px4-ros2-interface-lib/classpx4__ros2_1_1AttitudeSetpointType.html
   - Would be good to indicate what is experimental in these docs to reduce confusion - a note or similar on the experimental things
   - Would be good to make it clear in the modules section that this is not the only setpoint, just only the supported setpoint.
3. Make the name match the docs "Interface Library"
4. "Here is a list of all modules:" - again make it clear what is in library and what it means not to be here but in the class list.

bkueng:
I added the rest of it as well: https://auterion.github.io/px4-ros2-interface-lib/modules.html

The 'Here is a list of all modules' I cannot change - at least not easily. But it should contain all the directly usable classes & methods.

PR #8, opened 2023-12-05
tests: external navigation interface

GuillaumeLaine:
### Unit Tests:
- Publish invalid measurements and check exceptions are thrown correctly.

Test with `./build/px4_ros2_cpp/px4_ros2_cpp_unit_tests --gtest_filter='*Interface*'`

### Integration Tests:
- Publish measurement via interface and check that PX4 confirms measurement fusion via estimator status flags.

### Interface/Example Tweaks
- Log error message only once to avoid spamming the user when they catch interface exception.
- Add exception handling to examples

## Control mode interface
- Rename mode integration tests `Tester` -> `ModesTest`

**Jira**: [APX4-3591](https://auterion.atlassian.net/jira/software/c/projects/APX4/boards/225?selectedIssue=APX4-3591)

[APX4-3591]: https://auterion.atlassian.net/browse/APX4-3591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/px4_ros2_cpp/test/integration/global_navigation.cpp
+++ b/px4_ros2_cpp/test/integration/global_navigation.cpp
@@ -1,0 +1,118 @@
+/****************************************************************************
+ * Copyright (c) 2023 PX4 Development Team.
+ * SPDX-License-Identifier: BSD-3-Clause
+ ****************************************************************************/
+
+/**
+ * Summary:
+ *  Tests that measurements sent via the interface are fused into PX4.
+ * - `GlobalPositionInterfaceTest` fixture:
+ *   - Test Cases:
+ *     1. FuseAll: Sends measurements with all fields set. Expects cs_aux_gpos to be set to true.
+ */
+
+
+#include <gtest/gtest.h>
+#include <rclcpp/rclcpp.hpp>
+#include <px4_msgs/msg/estimator_status_flags.hpp>
+#include <px4_ros2/navigation/experimental/global_position_measurement_interface.hpp>
+#include "util.hpp"
+
+using namespace std::chrono_literals;
+using px4_ros2::GlobalPositionMeasurement, px4_ros2::GlobalPositionMeasurementInterface;
+using px4_msgs::msg::EstimatorStatusFlags;
+
+class GlobalPositionInterfaceTest : public BaseTest
+{
+protected:
+  void SetUp() override
+  {
+    _node = initNode();
+
+    _global_navigation_interface = std::make_shared<GlobalPositionMeasurementInterface>(*_node);
+    ASSERT_TRUE(_global_navigation_interface->doRegister()) <<
+      "Failed to register GlobalPositionMeasurementInterface.";
+
+    // Subscribe to PX4 EKF estimator status flags
+    _subscriber = _node->create_subscription<EstimatorStatusFlags>(
+      "/fmu/out/estimator_status_flags", rclcpp::QoS(10).best_effort(),
+      [this](EstimatorStatusFlags::UniquePtr msg) {
+        _estimator_status_flags = std::move(msg);
+      });
+  }
+
+  void waitUntilFlagsReset()
+  {
+    auto start_time = _node->now();
+
+    // Wait for PX4 to stop fusing external measurements
+    while (_estimator_status_flags->cs_aux_gpos) {
+      rclcpp::spin_some(_node);
+      rclcpp::sleep_for(kSleepInterval);
+
+      const auto elapsed_time = _node->now() - start_time;
+
+      if (elapsed_time >= kTimeoutDuration) {
+        FAIL() << "Timeout while waiting for PX4 estimator flags to reset.";
+        break;
+      }
+    }
+
+    _estimator_status_flags.reset();
+  }
+
+  void waitForMeasurementUpdate(
+    std::unique_ptr<GlobalPositionMeasurement> measurement,
+    const std::function<bool(const EstimatorStatusFlags::UniquePtr &)> & is_fused_getter
+  )
+  {
+    auto start_time = _node->now();
+
+    // Wait for PX4 estimator flags to confirm proper measurement fusion
+    while (!_estimator_status_flags || !is_fused_getter(_estimator_status_flags)) {
+
+      // Send measurement
+      measurement->timestamp_sample = _node->get_clock()->now();
+      ASSERT_NO_THROW(
+        _global_navigation_interface->update(*measurement)
+      ) << "Failed to send position measurement update via GlobalPositionMeasurementInterface.";
+
+      rclcpp::spin_some(_node);
+
+      // Check timeout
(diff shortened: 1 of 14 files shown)


px4_ros2_cpp/test/integration/global_navigation.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  void sendGlobalPositionMeasurement(
+    std::unique_ptr<GlobalPositionMeasurement> measurement,
+    const std::string & test_subname = "")
+  {
+    ASSERT_NO_THROW(global_navigation_interface->update(*measurement)) <<
+      "Failed to send position measurement update via GlobalPositionMeasurementInterface.";
+
+    constexpr auto timeout_duration = std::chrono::milliseconds(200);
```
bkueng, line 48: If you place above `using namespace std::chrono_literals;`, you can do:
```suggestion
    const auto timeout_duration = 200_ms;
```

px4_ros2_cpp/test/unit/global_navigation.cpp
```diff
@@ -0,0 +1,145 @@
...
+      *node);
+  }
+
+  std::shared_ptr<rclcpp::Node> node;
+  std::shared_ptr<px4_ros2::GlobalPositionMeasurementInterface> global_navigation_interface;
+};
+
+// Tests interface when position measurement is empty
+TEST_F(GlobalPositionMeasurementInterfaceTest, MeasurementEmpty) {
+  GlobalPositionMeasurement * measurement;
```
bkueng, line 49: Why not just
```suggestion
  GlobalPositionMeasurement measurement{};
```

px4_ros2_cpp/test/integration/global_navigation.cpp
GuillaumeLaine: Won't clang tidy complain about doing a namespace using-directives in this case?

px4_ros2_cpp/test/unit/global_navigation.cpp
bkueng: You can use `measurement = {};` to reset it.

GuillaumeLaine:
FYI: build publish debian packages currently failing because the px4_msgs foxy release isn't up to date and doesn't have the latest aux_gpos changes

px4_ros2_cpp/test/integration/global_navigation.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+      // Send measurement
+      measurement->timestamp_sample = _node->get_clock()->now();
+      ASSERT_NO_THROW(
+        _global_navigation_interface->update(*measurement)
+      ) << "Failed to send position measurement update via GlobalPositionMeasurementInterface.";
+
+      rclcpp::spin_some(_node);
+
+      const auto current_time = std::chrono::steady_clock::now();
```
bkueng, line 73: Can you generally use the ros time? That gives us the option to use simulated time later on.

px4_ros2_cpp/test/integration/global_navigation.cpp
```diff
@@ -0,0 +1,112 @@
...
+      const auto elapsed_time = std::chrono::duration_cast<std::chrono::milliseconds>(
```
bkueng, line 74: Is the cast required here?
```suggestion
      const auto elapsed_time = current_time - start_time;
```

px4_ros2_cpp/test/integration/global_navigation.cpp
```diff
@@ -0,0 +1,112 @@
...
+      // Check timeout
+      if (elapsed_time >= _k_timeout_duration) {
+        ASSERT_NE(_estimator_status_flags, nullptr) <<
+          "Missing feedback from PX4: no estimator status flags published over /fmu/out/estimator_status_flags.";
+        EXPECT_TRUE(is_fused_getter(_estimator_status_flags)) <<
+          "Position measurement update was not fused into the PX4 EKF.";
+        break;
+      }
+
+      std::this_thread::sleep_for(_k_sleep_interval);
```
bkueng, line 87: Here too: `rclcpp::sleep_for`

px4_ros2_cpp/test/integration/global_navigation.cpp
```diff
@@ -0,0 +1,112 @@
...
+    resetFlags();
+  }
+
+  std::shared_ptr<rclcpp::Node> _node;
+  std::shared_ptr<GlobalPositionMeasurementInterface> _global_navigation_interface;
+  rclcpp::Subscription<EstimatorStatusFlags>::SharedPtr _subscriber;
+  EstimatorStatusFlags::UniquePtr _estimator_status_flags;
+
+  const std::chrono::seconds _k_timeout_duration = 60s;
+  const std::chrono::milliseconds _k_sleep_interval = 50ms;
```
bkueng, line 99:
```suggestion
  static const std::chrono::seconds kTimeoutDuration = 60s;
  static const std::chrono::milliseconds kSleepInterval = 50ms;
```

px4_ros2_cpp/test/integration/local_navigation.cpp
```diff
@@ -0,0 +1,163 @@
...
+      "/fmu/out/estimator_status_flags", rclcpp::QoS(10).best_effort(),
+      [this](EstimatorStatusFlags::UniquePtr msg) {
+        _estimator_status_flags = std::move(msg);
+      });
+  }
+
+  void resetFlags()
+  {
+    // Wait for PX4 to stop fusing external measurements
+    while (_estimator_status_flags->cs_ev_pos || _estimator_status_flags->cs_ev_hgt ||
```
bkueng, line 52: This should have some timeout

px4_ros2_cpp/test/integration/local_navigation.cpp
```diff
@@ -0,0 +1,163 @@
...
+          "Missing feedback from PX4: no estimator status flags published over /fmu/out/estimator_status_flags.";
+        EXPECT_TRUE(is_fused_getter(_estimator_status_flags)) <<
+          "Position measurement update was not fused into the PX4 EKF.";
+        break;
+      }
+
+      std::this_thread::sleep_for(_k_sleep_interval);
+    }
+
+    resetFlags();
```
bkueng, line 97: How long does that take? If it takes a while, you could also do it only in one of the tests.
You could also rename this to reflect a bit better what it does (something like `waitUntilFlagsReset()`)

GuillaumeLaine: Less than one second normally. But we would still need to wait between each test for the estimator flags to reset, otherwise they might affect the result of the next test

bkueng (commented):
Looks good, can you squash the last commit into the relevant previous one?

PR #10, opened 2023-12-06
CI: Upgrade ros-tooling/setup-ros

DanMesh:
Fixes CI failures due to Empy version.

- See https://github.com/ros-tooling/setup-ros/pull/632


bkueng (approved):
Nice, thanks

PR #11, opened 2023-12-07
Build debian packages in CI

ThomasDebrunner:
This builds debian packages for focal / foxy, jammy / humble and jammy / rolling both for arm64 and amd64 on release

### DIFF
--- a/.github/workflows/build-publish-debian-packages.yml
+++ b/.github/workflows/build-publish-debian-packages.yml
@@ -1,0 +1,72 @@
+name: Build publish debian packages
+
+on:
+  release:
+    types: [published]
+  pull_request:
+    branches:
+      - '*'
+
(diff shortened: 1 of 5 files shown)


markusachtelik (approved):

.github/workflows/build-publish-debian-packages.yml
```diff
@@ -0,0 +1,72 @@
...
+    strategy:
+      matrix:
+        platform: [linux/amd64, linux/arm64]
+        distro:
+        - ubuntu: jammy
+          ros2: humble
+        - ubuntu: jammy
+          ros2: rolling
+        - ubuntu: focal
+          ros2: foxy
```
bkueng, line 23: Foxy is EOL. Can you add Iron instead?

PR #18, opened 2024-01-17
tests: send correct drone attitude measurement

GuillaumeLaine:
Integration test was giving "Yaw unstable" warnings.
Pre-flight sim drone attitude points East not North, changed published attitude measurement accordingly.

### DIFF
--- a/px4_ros2_cpp/test/integration/local_navigation.cpp
+++ b/px4_ros2_cpp/test/integration/local_navigation.cpp
@@ -140,7 +140,8 @@ TEST_F(LocalPositionInterfaceTest, fuseEvVel) {
 
 TEST_F(LocalPositionInterfaceTest, fuseEvYaw) {
   auto measurement = std::make_unique<LocalPositionMeasurement>();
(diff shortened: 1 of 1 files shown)


px4_ros2_cpp/test/integration/local_navigation.cpp
```diff
@@ -140,7 +140,8 @@ TEST_F(LocalPositionInterfaceTest, fuseEvVel) {
 
 TEST_F(LocalPositionInterfaceTest, fuseEvYaw) {
   auto measurement = std::make_unique<LocalPositionMeasurement>();
-  measurement->attitude_quaternion = Eigen::Quaternionf {1.0F, 0.0F, 0.0F, 0.0F};
+  measurement->attitude_quaternion =
```
bkueng, line 143: Can you add a comment that it expects the vehicle to point east?

bkueng (commented):
Ah, good find

px4_ros2_cpp/test/integration/local_navigation.cpp
GuillaumeLaine: Added a comment, unfortunately I don't think this was the CI issue. It was only causing a warning but not causing the test to fail

PR #20, opened 2024-01-23
Add geometry and conversion utils

GuillaumeLaine:
### Moves utils from auterion_sdk
- Utils for commonly used geometry/math
- Utils for commonly used ROS2 PX4 conversions (ENU/NED, FRD/FLU)

### Changes
- templated methods
- `wrapAngleToPlusMinusPi` returns in [-pi, pi) instead of [-pi, pi] up to floating point precision
- rename `wrapAngleToPlusMinusPi` -> `wrapPi`
- Renamed `ned` to `frd` for thrust of attitude and rate setpoints
- Split auterion_sdk `utils.cpp` -> `{geometry.cpp, frame_conversion.cpp}`

### Adds
- `quaternionToPitch` (and `...Roll`, `...Yaw`) functions. Rationale: existing`quaternionToEuler` maps pitch to [-pi/2, pi/2] by convention but there are use cases for a pitch in [-pi, pi] i.e. accumulating delta angles to detect a flip.
- Utils unit tests
- Utils doxygen docs

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/utils/frame_conversion.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/utils/frame_conversion.hpp
@@ -1,0 +1,179 @@
+/****************************************************************************
+ * Copyright (c) 2024 PX4 Development Team.
+ * SPDX-License-Identifier: BSD-3-Clause
+ ****************************************************************************/
+
+#pragma once
+
+#include <Eigen/Eigen>
+#include <px4_ros2/utils/geometry.hpp>
+
+namespace px4_ros2
+{
+
+/**
+ * @brief Converts attitude from NED to ENU frame.
+ * Performs reference frame change and quaternion rotation s.t. the identity quaternion points along the x-axis within its reference frame.
+ *
+ * @param q_ned Attitude quaternion in NED frame.
+ * @return Attitude quaternion in ENU frame.
+ */
+template<typename Type>
+Eigen::Quaternion<Type> attitudeNedToEnu(const Eigen::Quaternion<Type> & q_ned)
+{
+  static constexpr Type kHalfSqrt2 = static_cast<Type>(0.7071067811865476);
+  const Eigen::Quaternion<Type> q_ned_to_enu {0.0, kHalfSqrt2, kHalfSqrt2, 0.0};  // 180 deg along X, -90 deg along Z
+  const Eigen::Quaternion<Type> q_yaw_minus_pi_2{kHalfSqrt2, 0.0, 0.0, -kHalfSqrt2};  // -90 deg along Z
+
+  return q_ned_to_enu * q_ned * q_yaw_minus_pi_2 * q_ned_to_enu.inverse();
+}
+
+/**
+ * @brief Converts attitude from ENU to NED frame.
+ * Performs reference frame change and quaternion rotation s.t. the identity quaternion points along the x-axis within its reference frame.
+ *
+ * @param q_enu Attitude quaternion in ENU frame.
+ * @return Attitude quaternion in NED frame.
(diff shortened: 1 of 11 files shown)


px4_ros2_cpp/include/px4_ros2/utils/geometry.hpp
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @brief Convert quaternion to yaw angle;
+ *
+ * @param q The input quaternion
+ * @return Yaw angle corresponding to the given quaternion in range [-pi, pi]
+*/
+float quaternionToYaw(const Eigen::Quaternionf & q);
+
+/**
+ * @brief Transforms a point from body frame to world frame given the yaw of the attitude quaternion.
```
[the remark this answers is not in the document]
GuillaumeLaine: I think that makes sense, will do

px4_ros2_cpp/include/px4_ros2/utils/frame_conversion.hpp
```diff
@@ -0,0 +1,180 @@
...
+ * @param yaw_ned Yaw angle in NED frame [rad].
+ * @return Yaw angle in ENU frame [rad], wrapped to [-pi, pi].
+ */
+template<typename T>
+static inline T yawNedToEnu(const T yaw_ned_rad)
+{
+  return wrapPi(static_cast<T>(M_PI / 2.0) - yaw_ned_rad);
+}
+template float yawNedToEnu<float>(const float);
+template double yawNedToEnu<double>(const double);
```
bkueng, line 58: Is there a reason why you added these?

px4_ros2_cpp/src/utils/frame_conversion.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <px4_ros2/utils/frame_conversion.hpp>
+
+namespace px4_ros2
+{
+
+template<typename Type>
+Eigen::Quaternion<Type> attitudeNedToEnu(const Eigen::Quaternion<Type> & q_ned)
+{
+  static constexpr double kHalfSqrt2 = 0.7071067811865476;
```
bkueng, line 14:
```suggestion
  static constexpr Type kHalfSqrt2 = static_cast<Type>(0.7071067811865476);
```

px4_ros2_cpp/include/px4_ros2/utils/frame_conversion.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * @brief Converts attitude from NED to ENU frame.
+ * Performs reference frame change and quaternion rotation s.t. the identity quaternion points along the x-axis within its reference frame.
+ *
+ * @param q_ned Attitude quaternion in NED frame.
+ * @return Attitude quaternion in ENU frame.
+ */
+template<typename Type>
+Eigen::Quaternion<Type> attitudeNedToEnu(const Eigen::Quaternion<Type> & q_ned);
```
bkueng, line 22: Since these are templates, you need to place all implementations to the header now.

px4_ros2_cpp/include/px4_ros2/utils/frame_conversion.hpp
GuillaumeLaine: No, I was confused thanks.
I was initially declaring the templates in the header and putting the template implementation and explicit types instantiation in the source file. Forgot to remove the type instantiating when moving inline implementations to the header.

But anyways as you mentioned I just learned that it's standard practice to have template methods be header only so I'll do that :+1:

bkueng (approved):
Nice work

PR #24, opened 2024-02-05
Add fw attitude example

KonradRudin:
Updated setpoint type attitude to update with euler angles instead of quaternions only.
Set small fixed attitude setpoint example for constant banking to show interface for FW

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/attitude.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/attitude.hpp
@@ -27,7 +27,7 @@ public:
   float desiredUpdateRateHz() override {return 200.f;}
 
   void update(
-    const Eigen::Quaternionf & attidude_setpoint,
+    const Eigen::Quaternionf & attitude_setpoint,
     const Eigen::Vector3f & thrust_setpoint_frd, float yaw_sp_move_rate_rad_s = 0.f);
 
 private:
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/attitude.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/attitude.hpp
@@ -30,6 +30,20 @@ public:
     const Eigen::Quaternionf & attitude_setpoint,
     const Eigen::Vector3f & thrust_setpoint_frd, float yaw_sp_move_rate_rad_s = 0.f);
 
+  /**
+   * @brief Send an attitude setpoint to the flight controller.
+   * Euler angles follow the RPY extrinsic Tait-Bryan convention (equiv. YPR intrinsic)
+   *
+   * @param roll Attitude setpoint roll angle [rad]
+   * @param pitch Attitude setpoint pitch angle [rad]
+   * @param yaw Attitude setpoint yaw angle [rad]
+   * @param thrust_setpoint_body Throttle demand [-1, 1]^3
+   * @param yaw_sp_move_rate_rad_s Yaw setpoint move rate [rad/s]
+  */
+  void update(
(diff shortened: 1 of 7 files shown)


px4_ros2_cpp/src/control/setpoint_types/experimental/attitude.cpp
```diff
@@ -37,6 +37,40 @@ void AttitudeSetpointType::update(
...
+  px4_msgs::msg::VehicleAttitudeSetpoint sp{};
+  sp.timestamp = _node.get_clock()->now().nanoseconds() / 1000;
+
+  sp.roll_body = roll;
+  sp.pitch_body = pitch;
+  sp.yaw_body = yaw;
+
+  sp.yaw_sp_move_rate = yaw_sp_move_rate_rad_s;
+
+  Eigen::Quaternionf att_setpoint_q{Eigen::AngleAxisf(roll, Eigen::Vector3f::UnitX()) *
```
bkueng, line 58: You can use the util methods like here: https://github.com/Auterion/px4-ros2-interface-lib/pull/23/files#diff-73bebd6da924b3ed229ab58e3a1507dd06b157139b1b5afa52a4f41f6d6c6d58R175

bkueng (commented):
Cool, good to see some FW usage!
@Jaeyoung-Lim fyi

px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/attitude.hpp
```diff
@@ -27,9 +27,13 @@ class AttitudeSetpointType : public SetpointBase
   float desiredUpdateRateHz() override {return 200.f;}
 
   void update(
-    const Eigen::Quaternionf & attidude_setpoint,
+    const Eigen::Quaternionf & attitude_setpoint,
     const Eigen::Vector3f & thrust_setpoint_frd, float yaw_sp_move_rate_rad_s = 0.f);
 
+  void update(
```
GuillaumeLaine, line 33:
```suggestion
  /**
   * @brief Send an attitude setpoint to the flight controller.
   * Euler angles follow the RPY extrinsic Tait-Bryan convention (equiv. YPR intrinsic)
   *
   * @param roll Attitude setpoint roll angle [rad]
   * @param pitch Attitude setpoint pitch angle [rad]
   * @param yaw Attitude setpoint yaw angle [rad]
   * @param thrust_setpoint_body Throttle demand [-1, 1]^3
   * @param yaw_sp_move_rate_rad_s Yaw setpoint move rate [rad/s]
  */
  void update(
```

I think it might be worth documenting this method overload to specify which euler angle conventions we use to avoid confusion.

If we want to be consistent with the PX4 matrix lib, it uses [3-2-1 intrinsic Tait-Bryan](https://github.com/PX4/PX4-Autopilot/blob/8dae9905aa4ac294dca9eead9459cfddb9094072/src/lib/matrix/matrix/Euler.hpp#L7) euler angles i.e. the rotation order is yaw, pitch, roll and the axes of rotations change with each consecutive rotation (equivalent to 1-2-3 extrinsic).

Using Eigen, this would be:
```
  Eigen::Quaternionf att_setpoint_q{
    Eigen::AngleAxisf(yaw, Eigen::Vector3f::UnitZ()) *
    Eigen::AngleAxisf(pitch, Eigen::Vector3f::UnitY()) *
    Eigen::AngleAxisf(roll, Eigen::Vector3f::UnitX())};
```
(as Beat mentioned, identical to the util function being added)

KonradRudin:
Converted to draft, since it depends now on other PR https://github.com/Auterion/px4-ros2-interface-lib/pull/23 which should be merged first

bkueng:
https://github.com/Auterion/px4-ros2-interface-lib/pull/23 got merged

bkueng (approved):

PR #26, opened 2024-02-07
tests: add rad deg conversion unit tests

GuillaumeLaine:


px4_ros2_cpp/test/unit/utils/geometry.cpp
```diff
@@ -8,6 +8,22 @@
 #include <gtest/gtest.h>
 #include <px4_ros2/utils/geometry.hpp>
 
+using namespace px4_ros2::literals;
```
bkueng, line 11: clang-tidy is a bit overly cautious here
```suggestion
using namespace px4_ros2::literals; // NOLINT
```

GuillaumeLaine:
I just learned that the static, inline, and const qualifiers I added are redundant because the functions are header-only templates and we're passing the argument by value. I can remove them if you think they're unnecessary

bkueng:
> I just learned that the static, inline, and const qualifiers I added are redundant because the functions are header-only templates and we're passing the argument by value. I can remove them if you think they're unnecessary

Since you mention it, yes I would (and the reason why I didn't add them in the first place)

bkueng (approved):

PR #27, opened 2024-02-12
trajectory setpoint: finish update method

GuillaumeLaine:
- `TrajectorySetpointType` additionally updates position, acceleration, yaw and yaw_rate

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/trajectory.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/trajectory.hpp
@@ -25,7 +25,11 @@ public:
 
   Configuration getConfiguration() override;
 
-  void update(const Eigen::Vector3f & velocity_setpoint_ned_m_s);
+  void update(
+    const Eigen::Vector3f & velocity_ned_m_s,
+    const std::optional<Eigen::Vector3f> & acceleration_ned_m_s2 = {},
+    std::optional<float> yaw_ned_rad = {},
+    std::optional<float> yaw_rate_ned_rad_s = {});
 
 private:
   rclcpp::Node & _node;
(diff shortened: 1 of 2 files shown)


px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/trajectory.hpp
```diff
@@ -25,7 +25,11 @@ class TrajectorySetpointType : public SetpointBase
 
   Configuration getConfiguration() override;
 
-  void update(const Eigen::Vector3f & velocity_setpoint_ned_m_s);
+  void update(
+    const Eigen::Vector3f & position_ned_m,
+    const Eigen::Vector3f & velocity_ned_m_s,
+    const Eigen::Vector3f & acceleration_ned_m_s2, float yaw_ned_rad,
+    float yaw_rate_ned_rad_s);
```
bkueng, line 32: Can you follow the same pattern as we have for the goto setpoint?
```suggestion
    const std::optional<Eigen::Vector3f> & acceleration_ned_m_s2={}, std::optional<float> yaw_ned_rad = {},
    std::optional<float> yaw_rate_ned_rad_s = {});
```

examples/cpp/modes/mode_with_executor/include/mode.hpp
```diff
@@ -47,7 +47,10 @@ class FlightModeTest : public px4_ros2::ModeBase
 
     const float elapsed_s = (now - _activation_time).seconds();
     const Eigen::Vector3f velocity{10.f, elapsed_s * 2.f, -2.f};
-    _trajectory_setpoint->update(velocity);
+    const Eigen::Vector3f acceleration{NAN, NAN, NAN};
+    const float yaw = NAN;
+    const float yaw_rate = NAN;
```
bkueng, line 52: You can remove those now and use the defaults (below too)

GuillaumeLaine: ah good point, thanks

bkueng (approved):

PR #28, opened 2024-02-14
Interrupting an existing flight mode by a custom flight mode executor

bastianhjaeger:
@bkueng 
I consider this a bad example and not to merge this into the repo, but rather a base of conversation. 

While looking at the interface lib, I thought about a possibility to run a flight mode executor, which triggers another flight mode, e.g. mission. Now consider an sensor based event to tell PX4 to switch from mission mode to the owned flight mode. 

This example is doing this (timer based).

Question: 
Was this use case considered and is there yet a better solution to it? 
(Maybe some way to inherit all mission flight mode functionally and add some stuff on top?)

### Usage: 

```
colcon build --packages-select example_mode_with_interrupting_executor_cpp
. install/setup.bash
ros2 run example_mode_with_interrupting_executor_cpp example_mode_with_interrupting_executor_cpp
```

Create a small mission on AMC and upload it.

Then start the flight mode.


bkueng:
> Was this use case considered and is there yet a better solution to it?
(Maybe some way to inherit all mission flight mode functionally and add some stuff on top?)

Yes I expect there to be mode executors with more complex mode switching logic. You're generally free to do what you want here.
I think the main part that's still missing is mission execution (e.g. planning & executing a trajectory from a mode) and GCS integration (allowing a user to configure a mode what it should do, which requires the GCS to know its capabilities & providing visual feedback - this is a big topic).

PR #29, opened 2024-02-16
Local global conversion

GuillaumeLaine:


px4_ros2_cpp/src/utils/geodesic.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+namespace px4_ros2
+{
+
+MapProjection::MapProjection(Context & context)
+: _node(context.node())
+{
+  _map_projection_math = std::make_shared<MapProjectionImpl>();
+  _vehicle_local_position_sub = _node.create_subscription<px4_msgs::msg::VehicleLocalPosition>(
+    "/fmu/out/vehicle_local_position", rclcpp::QoS(10).best_effort(),
```
bkueng, line 18: `context.topicNamespacePrefix() + "/fmu/out`
`rclcpp::QoS(10).best_effort(),` -> `rclcpp::QoS(1).best_effort(),`

px4_ros2_cpp/src/utils/geodesic.cpp
```diff
@@ -0,0 +1,185 @@
...
+    [this](px4_msgs::msg::VehicleLocalPosition::UniquePtr msg) {
+      const uint64_t timestamp = msg->ref_timestamp;
+      if (!isInitialized() && std::isfinite(timestamp) && timestamp != 0) {
```
bkueng, line 21: `std::isfinite(timestamp)` is always true

px4_ros2_cpp/src/utils/geodesic.cpp
```diff
@@ -0,0 +1,185 @@
...
+        _map_projection_math->initReference(
+          msg->ref_lat, msg->ref_lon, msg->ref_alt,
+          msg->ref_timestamp);
```
bkueng, line 24: The reference might change later on (on ekf resets). You can use `ref_timestamp` to keep track of that. We should also add a callback, so a user can get notified when that happens.

px4_ros2_cpp/include/px4_ros2/control/setpoint_types/goto.hpp
```diff
@@ -53,5 +54,37 @@ class GotoSetpointType : public SetpointBase
...
+    const Eigen::Vector3d & global_position,
+    const std::optional<float> & heading = {},
+    const std::optional<float> & max_horizontal_speed = {},
+    const std::optional<float> & max_vertical_speed = {},
+    const std::optional<float> & max_heading_rate = {}
+  );
+
+private:
+  rclcpp::Node & _node;
+  std::shared_ptr<MapProjection> _map_projection;
```
bkueng, line 85: These can also be `unique_ptr`

px4_ros2_cpp/src/control/setpoint_types/goto.cpp
```diff
@@ -64,4 +64,34 @@ SetpointBase::Configuration GotoSetpointType::getConfiguration()
...
+
+void GotoGlobalSetpointType::update(
+  const Eigen::Vector3d & global_position,
+  const std::optional<float> & heading,
+  const std::optional<float> & max_horizontal_speed,
+  const std::optional<float> & max_vertical_speed,
+  const std::optional<float> & max_heading_rate)
+{
+  if (!_map_projection->isInitialized()) {
+    RCLCPP_ERROR(
```
bkueng, line 85: `RCLCPP_ERROR_ONCE`

examples/cpp/modes/goto_global/include/mode.hpp
```diff
@@ -0,0 +1,224 @@
...
+    _goto_setpoint = std::make_shared<px4_ros2::GotoGlobalSetpointType>(*this);
+
+    _vehicle_global_position = std::make_shared<px4_ros2::OdometryGlobalPosition>(*this);
+
+    _vehicle_attitude_sub = node.create_subscription<px4_msgs::msg::VehicleAttitude>(
+      "/fmu/out/vehicle_attitude", rclcpp::QoS(1).best_effort(),
+      [this](px4_msgs::msg::VehicleAttitude::UniquePtr msg) {
+        updateVehicleHeading(msg);
+      }
+    );
```
bkueng, line 38: We should have some odometry class for attitude (we can add that later on)

examples/cpp/modes/goto_global/include/mode.hpp
```diff
@@ -0,0 +1,224 @@
...
+    _mode = std::make_unique<FlightModeTest>(*this);
+
+    if (!_mode->doRegister()) {
+      throw std::runtime_error("Registration failed");
+    }
+  }
+
+private:
+  std::unique_ptr<FlightModeTest> _mode;
+};
```
bkueng, line 224: We can start to remove this boiler-plate by adding a templated class for this, in the form of:
```
template<typename ModeT>
class NodeWithMode : public rclcpp::Node
{
     NodeWithMode(string node_name, bool enable_debug_output=false)
```
(and even better if you find a better name :)

bkueng (commented):
Nice, looks very clean.
Potentially add some unit tests for the geodesic methods.

examples/cpp/modes/goto_global/include/mode.hpp
GuillaumeLaine: Great point, I'll make a PR for that after this one

px4_ros2_cpp/include/px4_ros2/utils/geodesic.hpp
```diff
@@ -0,0 +1,250 @@
...
+  Eigen::Vector3d localToGlobal(const Eigen::Vector3f & local_position) const;
+
+private:
+  /**
+   * @throw runtime error if class instance is not initalized
+   */
+  void assertInitalized() const;
+
+  rclcpp::Node & _node;
+  std::shared_ptr<MapProjectionImpl> _map_projection_math;
```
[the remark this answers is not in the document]
GuillaumeLaine: I initially struggled to do this until I found this thread: https://stackoverflow.com/questions/13414652/forward-declaration-with-unique-ptr. I'm not sure I fully get it and you may have already known, thought I'd share just in case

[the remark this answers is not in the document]
bkueng: Ah, no, I was not aware of that.

px4_ros2_cpp/src/utils/geodesic.cpp
bkueng: Yes this condition. It might happen during normal operation, so I would not warn.
The additional callback is used so the user can update other local references. This is where it gets complicated and we'll need to think a bit about it. We can also do that later.

examples/cpp/modes/goto_global/include/mode.hpp
GuillaumeLaine: I'm creating a [separate PR](https://github.com/Auterion/px4-ros2-interface-lib/pull/30) for this so that there's more room for discussion, also since it's not directly related to this global/local conversion PR.

GuillaumeLaine:
Thanks for the review, I'm done applying your suggestions for this PR. I'll open separate PRs for the suggestions that weren't directly related to this PR:
- Odometry class for attitude
- Node with Mode class
- Callback for ekf resets in the `MapProjection` class

bkueng (approved):

PR #31, opened 2024-02-26
Add OdometryAttitude class

GuillaumeLaine:
### Changes
- Add `OdometryAttitude` class.
- Add `heading()` getter to `OdometryLocalPosition` class.
- Use these interfaces in examples.

### Note
`OdometryAttitude::yaw()` and `OdometryLocalPosition::heading()` return the same value (up to PX4 timing), but I think it still makes sense to add a `OdometryLocalPosition::heading()` method for cases where the local position is already subscribed to e.g. in our local goto example.

**JIRA**: [APX4-3816](https://auterion.atlassian.net/browse/APX4-3816)

[APX4-3816]: https://auterion.atlassian.net/browse/APX4-3816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/odometry/attitude.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/odometry/attitude.hpp
@@ -1,0 +1,71 @@
+/****************************************************************************
+ * Copyright (c) 2024 PX4 Development Team.
+ * SPDX-License-Identifier: BSD-3-Clause
+ ****************************************************************************/
+
+#pragma once
+
+#include <Eigen/Eigen>
+#include <px4_msgs/msg/vehicle_attitude.hpp>
+#include <px4_ros2/common/context.hpp>
+#include <px4_ros2/odometry/subscription.hpp>
+#include <px4_ros2/utils/geometry.hpp>
+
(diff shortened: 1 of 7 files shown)


px4_ros2_cpp/include/px4_ros2/odometry/attitude.hpp
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * @brief Get the vehicle's roll in extrinsic RPY order.
+   *
+   * @return the attitude roll in radians within [-pi, pi]
+   */
+  float roll() const
+  {
+    const px4_msgs::msg::VehicleAttitude & att = last();
+    const Eigen::Quaternionf q{att.q[0], att.q[1], att.q[2], att.q[3]};
+    return quaternionToRoll(q);
```
bkueng, line 48: You could simplify these a bit:
```suggestion
    return quaternionToRoll(attitude());
```

bkueng (approved):

PR #36, opened 2024-05-03
gh-actions: update distro to noble for ROS2 rolling

GuillaumeLaine:


bkueng:
Can you have a look at the CI failure? https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/8939779841/job/24556494270?pr=36

GuillaumeLaine:
Found an answer on this [thread](https://stackoverflow.com/questions/72871100/compiler-library-or-user-error-eigenarray-gcc-12-1-array-subscript) on Stack Overflow: it seems like since gcc 12.1 this warning occurs under specific conditions when the compiler incorrectly asses the size of Eigen objects being assigned values using `.head()` or `.tail()`.

And gcc is updated 11 -> 13 from Ubuntu 22.04 -> 24.04.

bkueng (approved):

PR #37, opened 2024-05-10
control: reduce default control update frequency

GuillaumeLaine:
### Changes
- Reduce default control update frequency to 100Hz
- Do not trigger `setSetpointUpdateRateFromSetpointTypes()` after registration if the mode has already set a desired setpoint update rate (e.g. already called `setSetpointUpdateRate()` in the mode constructor)

### Testing
- Tested in SITL that the expected control frequency is set when specifying / not specifying an update rate in the mode's constructor / onActivate

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp
@@ -31,7 +31,7 @@ public:
   ~DirectActuatorsSetpointType() override = default;
 
   Configuration getConfiguration() override;
-  float desiredUpdateRateHz() override {return 100.f;}
+  float desiredUpdateRateHz() override {return 200.f;}
 
   /**
    * Send servos setpoint
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp
@@ -31,7 +31,7 @@ public:
   ~DirectActuatorsSetpointType() override = default;
 
   Configuration getConfiguration() override;
-  float desiredUpdateRateHz() override {return 500.f;}
+  float desiredUpdateRateHz() override {return 100.f;}
 
   /**
    * Send servos setpoint
--- a/px4_ros2_cpp/src/components/mode.cpp
+++ b/px4_ros2_cpp/src/components/mode.cpp
@@ -229,6 +229,7 @@ bool ModeBase::onRegistered()
 
   activateSetpointType(*_setpoint_types[0]);
   if (_setpoint_update_rate_hz < FLT_EPSILON) {
+    // Do not use default setpoint rate if rate was already set by user
     setSetpointUpdateRateFromSetpointTypes();
   }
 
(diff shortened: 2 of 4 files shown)


px4_ros2_cpp/src/components/mode.cpp
```diff
@@ -228,7 +228,9 @@ bool ModeBase::onRegistered()
   // TODO: check setpoint types compatibility with current vehicle type
 
   activateSetpointType(*_setpoint_types[0]);
-  setSetpointUpdateRateFromSetpointTypes();
+  if (_setpoint_update_rate_hz < FLT_EPSILON) {
```
bkueng, line 231: Can you add a comment that we don't update if the user already set it?

px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp
```diff
@@ -31,7 +31,7 @@ class DirectActuatorsSetpointType : public SetpointBase
   ~DirectActuatorsSetpointType() override = default;
 
   Configuration getConfiguration() override;
-  float desiredUpdateRateHz() override {return 500.f;}
+  float desiredUpdateRateHz() override {return 100.f;}
```
bkueng, line 34: I don't think these should be lowered. At least not by that much. Direct motor control & rate controller on an MC needs to run fast enough to work correctly.

bkueng (changes requested):

px4_ros2_cpp/include/px4_ros2/control/setpoint_types/direct_actuators.hpp
GuillaumeLaine: @markusachtelik was advocating for 100Hz defaults, what is the decision here?

markusachtelik: 500 Hz gets the CPU up by 20-30%. in addition, something seems to queue up and px4 says the mode becomes unresponsive. 
Rate control is definitely enough at 100 Hz, we can debate 200 Hz for direct motor control. If someone really knows what he/she is doing, the update rate can still be increased. But the default should be something not crazy.

bkueng: The CPU load depends on various factors. I used a rate-controlled mode with this code on Skynode, and didn't have any problems.
I'm fine with lowering it a bit, but 100+200Hz are too low for the general case.

markusachtelik: everything faster than 200Hz (5ms) will not work well on the scheduler on a non RT OS. let's please set this up so it works in the majority of cases by default. If someone wants it faster, there is still the setFrequency() call, which should work now. 

can we get this merged please?  

(Anecdotal evidence: My team and I have done rate and direct motor controllers with 100 Hz update rate in the past, the latter even flew a flip. Even the quadcopter dynamics are not as crazy as we always think. maybe a very small tiny quad.)

GuillaumeLaine: I've just set direct actuators and rates to 200Hz default. Hopefully this is compromising enough, someone feel free to merge once the CI is green (:

bkueng (approved):

PR #39, opened 2024-06-04
fix usage of ROS namespaces

damien-robotsix:
Currently the library do not take into account the namespace if specified. The micro DDS client allows in PX4 to configure a custom namespace (multiple drones on the same network). To reflect this, it would be good to also allow this library to use custom namespaces. Removing "/" in front of topic definitions solve this. The topics will reflect the chosen namespace for the node.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": false
+}
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Compatibility is only guaranteed if using latest `main` on the PX4 and px4_ros2/
 The library checks for message compatibility on startup when registering a mode.
 `ALL_PX4_ROS2_MESSAGES` defines the set of checked messages. If you use other messages, you can check them using:
 ```cpp
-if (!px4_ros2::messageCompatibilityCheck(node, {{"/fmu/in/vehicle_rates_setpoint"}})) {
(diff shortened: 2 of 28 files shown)


bkueng:
There's already a `topic_namespace_prefix` for that. You can pass that to the `Mode`'s constructor.
That said I'm not sure whether that's the best solution though.

damien-robotsix:
Ho, my bad, I was too fast on this, I didn't see the topic prefix in the constructor.

However that breaks the ability to set the namespace using ROS arguments when running the mode Node.

Anyway bot approach are compatible. With this branch, you can:

- use "/" as prefix to avoid the node namespace to be propagated on the /fmu/...
- use "/blabla/" to enforce the namespace for the fmu topics
- use no prefix to get the node namespace propagated to the fmu topics

How does that sound?

bkueng:
> However that breaks the ability to set the namespace using ROS arguments when running the mode Node.

That is correct. It's separate, as someone might want to write a node that controls/consumes telemetry from multiple vehicles.

> Anyway both approaches are compatible. With this PR, you can:

I'm fine with both approaches, just the default needs to work with the PX4 default, independent from the node name. Which looks like it's the case here, right?

damien-robotsix:
It will if the node has no namespace (default in ROS2)

bkueng (approved):
Thanks for clarifying

damien-robotsix:
Thanks for merging :)

PR #42, opened 2024-06-27
fix registering issue when node is not immediately spinned

damien-robotsix:
Hello,

Here is a suggestion for the classes in NodeWithMode. They immediately trigger the registering but if you don't spin the node fast enough (like I do since I need extra configuration after the creation of the instance)
- then you quickly get flagged unresponsive by the Autopilot 
- then disconnection happen because the Autopilot is not sending you check request anymore when flagged unresponsive (is it wanted? or a bug?)

In all cases, it could be good to either:
1) let the hand to the user to trigger the registration just before spinning the node 
2) have a more complex mechanism to check that the node is spinning before doing the registration.

The modification if put in this PR is for solution 1). However, this changes usage as the user must do the registration "manually" before spinning the node. 
So I am open to suggestions on this.

Also, I did a minor modification to access the Executor because you might need that if you want to access specific functions you implemented in the class inherited from ExecutorBase.

Damien

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/components/node_with_mode.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/components/node_with_mode.hpp
@@ -50,7 +50,10 @@ public:
     }
 
     _mode = std::make_unique<ModeT>(*this);
+  }
 
+  void doRegister()
+  {
     if (!_mode->doRegister()) {
       throw std::runtime_error("Registration failed");
     }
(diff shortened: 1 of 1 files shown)


bkueng:
I think in your case I would rather not use `NodeWithMode`, and instead manually create the ros node, which gives you more flexibility.
`NodeWithMode` is meant for simplicity, and adding an extra registration call would not allow to use it like this anymore:
`rclcpp::spin(std::make_shared<px4_ros2::NodeWithMode<MyExecutor, MyMode>>("my_node"));`

PR #43, opened 2024-06-27
add angular velocity feedback

damien-robotsix:
Hello,

I'd like to add the angular velocity to the odometry messages. It is sad to have almost all the odometry but this.

It doesn't work with the current master branch in PX4-Autopilot, the message is missing in the DDS Yaml configuration file. So it need to be updated also there.

Thanks


bkueng (approved):
Thanks, there's definitely more telemetry we can add.

PR #47, opened 2024-07-22
wait for fmu before mode registration

GuillaumeLaine:


px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -50,7 +51,9 @@ bool ModeExecutorBase::doRegister()
 
   assert(!_registration->registered());
 
-  if (!messageCompatibilityCheck(_node, {ALL_PX4_ROS2_MESSAGES}, _topic_namespace_prefix)) {
+  if (!waitForFMU(node(), 15s) ||
```
GuillaumeLaine, line 54: The mode executor could check if its owned mode is set to skip the message check?
This would require adding a getter to the `Mode` class though

bkueng: It's only used for unit tests

PR #48, opened 2024-07-23
mode: extends available replacements to precision land mode

GuillaumeLaine:


px4_ros2_cpp/include/px4_ros2/components/mode.hpp
```diff
@@ -82,6 +82,8 @@ class ModeBase : public Context
     px4_msgs::msg::VehicleStatus::NAVIGATION_STATE_AUTO_LAND;
   static constexpr ModeID kModeIDRtl =
     px4_msgs::msg::VehicleStatus::NAVIGATION_STATE_AUTO_RTL;
+  static constexpr ModeID kModeIDPrecLand =
```
bkueng, line 85:
```suggestion
  static constexpr ModeID kModeIDPrecisionLand =
```

bkueng (approved):

PR #49, opened 2024-07-23
distanceGround() and updatePosition() using TrajectorySetpoint

dakejahl:


px4_ros2_cpp/src/control/setpoint_types/experimental/trajectory.cpp
```diff
@@ -42,6 +42,12 @@ void TrajectorySetpointType::update(
 
 }
 
+void TrajectorySetpointType::update(const px4_msgs::msg::TrajectorySetpoint& setpoint)
```
bkueng, line 45: Which parts are you missing from the above interface? Position? jerk?

px4_ros2_cpp/include/px4_ros2/odometry/local_position.hpp
```diff
@@ -71,6 +71,12 @@ class OdometryLocalPosition : public Subscription<px4_msgs::msg::VehicleLocalPos
     const px4_msgs::msg::VehicleLocalPosition & pos = last();
     return pos.heading;
   }
+
+  float distanceGround() const
```
bkueng, line 75: Can you please separate the commit?

px4_ros2_cpp/src/control/setpoint_types/experimental/trajectory.cpp
dakejahl: Yes I want set the entire message directly

bkueng: Can you rather add another `update` method that allows for that? A method is just easier to use, and makes it clearer which fields are optional.

dakejahl: IMO a method with 6 arguments is harder to use than a method with a single struct. The only advantage I can see for using a method with std::optional is to force the implementation to provide nullopt which ensures those values are NAN, rather than the user needing to remember to set those fields to NAN before passing the struct. Is that the reason you prefer it?

bkueng: Yes and a few other factors (the message contains other fields that do not need to be set (timestamp), having to convert from eigen vectors and API consistency).
For the too-many-arguments case I'd rather introduce an intermediate struct that allows to use Eigen vectors and std::optional.

dakejahl:
Something else I've noticed is that you cannot control both position and velocity simultaneously. For precision land for example you want to control XY position and Z velocity but when I send a Z velocity setpoint PX4 no longer controls the XY position.

dakejahl:
@bkueng thoughts?

patrikpordi:
Is there any progress with this PR?

px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/trajectory.hpp
```diff
@@ -36,6 +36,9 @@ class TrajectorySetpointType : public SetpointBase
     std::optional<float> yaw_ned_rad = {},
     std::optional<float> yaw_rate_ned_rad_s = {});
 
+  void updatePosition(
```
bkueng, line 39: For position only, there's already a goto setpoint: https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/goto.hpp
Does that not fit your use-case?

dakejahl: The go-to setpoint did not work as well. I am also controlling velocity in my application.

bkueng: Ok I don't mind adding this, but please add a comment to generally prefer the goto setpoint. It's also possible to switch between multiple setpoint types in the same mode.

There's also some CI failures (you can rebase for the build failure).

dakejahl:
@bkueng I rebased and added a comment to the function declaration

PR #52, opened 2024-08-07
Python pre commit

GuillaumeLaine:


bkueng (commented):
I was hesitant in the past to use black as it's very intrusive. We can give it a try though.
I generally rely on the help of the IDE to reformat too long lines.

PR #55, opened 2024-08-23
log: descriptive timeout messages

GuillaumeLaine:


px4_ros2_cpp/src/components/wait_for_fmu.cpp
```diff
@@ -47,7 +47,7 @@ bool waitForFMU(
       }
 
     } else {
-      RCLCPP_DEBUG(node.get_logger(), "timeout");
+      RCLCPP_DEBUG(node.get_logger(), "timeout while waiting for fmu");
```
DanMesh, line 50:
```suggestion
      RCLCPP_DEBUG(node.get_logger(), "timeout while waiting for FMU");
```

px4_ros2_cpp/src/components/message_compatibility_check.cpp
```diff
@@ -207,7 +207,7 @@ RequestMessageFormatReturn requestMessageFormat(
         }
 
       } else {
-        RCLCPP_INFO(node.get_logger(), "timeout");
+        RCLCPP_INFO(node.get_logger(), "timeout while checking message compatibility");
       }
     }
   }
```
DanMesh, line 213: Just above this we also have the same "no message received" logs in each of these three cases ([line](https://github.com/Auterion/px4-ros2-interface-lib/pull/55/files#diff-af20ef344ec015f612dd4bee7cc5e1f6c00df84bbec977b6376c7bd387a22621R206)). Maybe we can improve those too while you're at it 😃

DanMesh (approved):
A very necessary and important improvement - thanks!

px4_ros2_cpp/src/components/message_compatibility_check.cpp
GuillaumeLaine: Good point. I took me a minute but now done :D

PR #59, opened 2024-10-01
Add global navigation reset method

GuillaumeLaine:
### Changes

Add reset method to increment the position reset counter

Only incrementing the lat_lon reset counter since it seems to be the only counter processed in the AuxGlobalPosition module in PX4: https://github.com/Auterion/PX4_firmware_private/blob/eaa169d211a19ecbd37a454846eb117e40ec204d/src/modules/ekf2/EKF/aid_sources/aux_global_position/aux_global_position.cpp#L59

### Testing

Only unit-testing


bkueng (approved):

bkueng:
There's some CI failures: https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/11135419376/job/30945399106?pr=59

PR #72, opened 2025-01-17
fix mode_executor under ROS Jazzy: create vehicle command ack subscription right before use

bkueng:
The behavior between ROS Humble and Jazzy differs: Jazzy now throws an exception 'subscription already associated with a wait set'. Apparently because the subscription is already in the node executor. This works around that by creating the subscription right before use.

This also adds retries when a command/ack times out.

Tested on Jazzy and Humble.


henrykotze:
Can confirmed it resolves the 'subscription already associated with a wait set' exception. 

Tested on Jazzy

bkueng:
Thanks for testing

PR #73, opened 2025-01-27
updated the script to consider versioned paths

Claudio-Chies:
Currently, the script always failed because it can't find the files inside the versioned message folder, with this commit, I'm searching all the subfolders in the `msg` directory, to find the files. Thereby, messages in the versioned folder are also found.
An open issue is that the `px4_msgs` are not up-to-date with the version field in the message definition, and thereby the check fails.
There, the question would be if that field will also be present in the px4_msgs repo, or should we just exclude this in the compensation? @bkueng

### DIFF
--- a/scripts/check-message-compatibility.py
+++ b/scripts/check-message-compatibility.py
@@ -19,7 +19,7 @@ def message_fields_str_for_message_hash(topic_type: str, msgs_dir: str) -> str:
     Reads the .msg file corresponding to the given topic type, extracts field definitions,
     and recursively processes nested types to generate a string representation of all fields.
     """
-    filename = f"{msgs_dir}/msg/{topic_type}.msg"
+    filename = find_file_in_subfolders(f"{msgs_dir}/", f"{topic_type}.msg")
     try:
         with open(filename, 'r') as file:
             text = file.read()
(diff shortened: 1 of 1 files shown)


scripts/check-message-compatibility.py
```diff
@@ -151,14 +158,14 @@ def main(repo1: str, repo2: str, verbose: bool = False):
     else:
         if verbose:
             for msg_type in incompatible_types:
-                file1 = os.path.join(repo1, 'msg', f'{msg_type}.msg')
-                file2 = os.path.join(repo2, 'msg', f'{msg_type}.msg')
+                file1 = find_file_in_subfolders(repo1, f'{msg_type}.msg')
```
bkueng, line 161: I'd keep the `msg` part to ensure the same file is found as above
```suggestion
                file1 = find_file_in_subfolders(os.path.join(repo1, 'msg'), f'{msg_type}.msg')
```

bkueng (commented):
> An open issue is that the px4_msgs are not up-to-date with the version field in the message definition, and thereby the check fails.
There, the question would be if that field will also be present in the px4_msgs repo, or should we just exclude this in the compensation? @bkueng

Are you sure it's causing a failed check? I'm asking because constants should explicitly be excluded here already: https://github.com/Auterion/px4-ros2-interface-lib/blob/9cc653945077dbba280da39b5a7a77d7b5aaf1ed/scripts/check-message-compatibility.py#L48

scripts/check-message-compatibility.py
```diff
@@ -114,12 +114,19 @@ def compare_files(file1: str, file2: str):
     with open(file1, 'r') as f1, open(file2, 'r') as f2:
         diff = list(difflib.unified_diff(f1.readlines(), f2.readlines(), fromfile=file1, tofile=file2))
         if diff:
-            print(f"Mismatch found between {file1} and {file2}:")
+            print(f"Mismatch found between {file1} and {file2}: ")
```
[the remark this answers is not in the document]
Claudio-Chies: i wasnt able to commit the change without because of the formatting requirements from the precommit hook.

scripts/check-message-compatibility.py
Claudio-Chies: fixed

Claudio-Chies:
> > An open issue is that the px4_msgs are not up-to-date with the version field in the message definition, and thereby the check fails.
> > There, the question would be if that field will also be present in the px4_msgs repo, or should we just exclude this in the compensation? @bkueng
> 
> Are you sure it's causing a failed check? I'm asking because constants should explicitly be excluded here already:
> 
> https://github.com/Auterion/px4-ros2-interface-lib/blob/9cc653945077dbba280da39b5a7a77d7b5aaf1ed/scripts/check-message-compatibility.py#L48

you are right, the version is excluded in the check, i updated the scripts so that they now work. there was an open issue with the path.

bkueng (approved):
Looks good, thanks

PR #74, opened 2025-01-30
apply version suffix to fmu topics

GuillaumeLaine:
### Changes
- Apply message version suffix to all FMU topics

### Testing
- Tested with example goto node built with a different version of `VehicleLocalPosition` than PX4 in SITL
- Translation node (tested after rebase too) implemented a direct translation between the two topic message version
- Messages transferred properly

### Note
Kept the message compatibility check as the default at the moment. Since the message translation node is not merged yet, I've marked the translation node option as "experimental" in the readme, and kept message matching as the default way to use the px4-ros2 lib for the time being

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # PX4 ROS 2 Interface Library
 
 Library to interface with PX4 from a companion computer using ROS 2.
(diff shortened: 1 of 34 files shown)


bkueng (approved):
Nice

PR #76, opened 2025-02-10
Add FW control API

mahimayoga:
Implements one FW control API that unifies `FWLongitudinalControlSetpoint.msg` and `FWLateralControlSetpoint.msg` as `FWControlSetpointType`.


px4_ros2_cpp/include/px4_ros2/control/setpoint_types/fixedwing/fw_control_setpoint.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+	/**
+	 * @brief Update
+	 *
+	 * Unset optional values are not controlled
+	 *
+	 */
+	void update(
+			    const float course_setpoint,
+				const float heading_setpoint,
```
RomanBapst, line 42: @mahimayoga Let's call this `airspeed_reference_direction` to be very precise. I also named it that way in PX4.

RomanBapst:
image
@mahimayoga I think the CI failure is real, probably need to get rid of the fw global goto stuff

mahimayoga:
@RomanBapst I think the issue is that in the test there is a dependency on px4_msgs. 
image
 

(https://github.com/Auterion/px4-ros2-interface-lib/blob/main/dependencies.repos)

PR #77, opened 2025-02-13
Node with mode executor and multiple modes

Alefunk:
I added a variadic template parameter to the `NodeWithModeExecutor` to enable registering of multiple custom modes with one node and easily access their ids from within the executor. The behavior of the class has not changed if just one mode is passed as template parameter. For each additional mode a unique_ptr of the optional modes will be generated and saved in a tuple. All modes of the tuple will be registered together with the executor. The executor itself must take the additional modes as constructor arguments, as is already the case with the owned mode.

The provided example (`example_executor_with_multiple_modes`) has two additional modes in comparison to the mode-with-executor example, so three custom modes in total:

1. Auto-Executor-Start doing the same as in the mode-with-executor example. This mode is owned by the executor.
2. Auto-Executor-Segment doing a flip with roll setpoints.
3. Auto-Executor-End flying in a zig-zag pattern.

### DIFF
--- a/examples/cpp/modes/executor_with_multiple_modes/CMakeLists.txt
+++ b/examples/cpp/modes/executor_with_multiple_modes/CMakeLists.txt
@@ -1,0 +1,34 @@
+cmake_minimum_required(VERSION 3.5)
+project(example_executor_with_multiple_modes_cpp)
+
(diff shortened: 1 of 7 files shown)


bkueng (approved):
Nice, thanks

Issue #78, opened 2025-02-13
Why Trajectory setpoint is split into two parts?


benchaso:
The trajectory setpoint is split into two part "update" and "updatePosition"?

This make not possible to send position and yaw at the same time. So if wanted to do yaw tracking by keeping the same position is not possible?

bkueng:
Because position and velocity should not be used together generally to prevent inconsistencies.
It sounds like you better use the [goto setpoint](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/goto.hpp) for your use-case.

benchaso:
Ok, thanks. In fact, goto setpoint could be better

azerupi:
>Because position and velocity should not be used together generally to prevent inconsistencies.

Does this apply if we control individual axis? For example control x, y in velocity but z in position.

Isn't altitude mode in PX4 implemented with position + acceleration setpoints? The current API doesn't seem like it would allow us to implement our own altitude mode. Is that correct?

azerupi:
@bkueng thanks for the clarification!

>And you can also define your own setpoint type class (e.g. by copying TrajectorySetpointType) in your code and then send any of the PX4 setpoint msg types.

That's a good point, I was fixated in the API of the library that I forgot we could subclass the type and extend it however we want. If this is a commonly expected pattern it might be a good idea to call it out in the PX4 docs on external modes?

>Generally, yes PX4 is making use of that, but also makes sure it's not sending inconsistent setpoints. I'm fine with exposing the full set, but it needs a clear 'you need to know what you are doing' kind of comment.

If I were to send a PR in that direction, how would you prefer the interface to look like?

1. Overload the `update` method to take in all parameters (position, velocity, acceleration, yaw, yaw rate)
2. Overload the `update` method to take the px4 message type directly
3. Define a new struct in the same way as `LocalPositionMeasurement` where all fields are wrapped in `std::optional` that is then passed to an overloaded `update` method? Anything that isn't set explicitly it set to `NaN`. Would we want to separate all axis into their own optional so that one can set position only on one axis without the risk of setting the other two axis to 0 instead of `Nan`
4. Something else?

bkueng:
> If this is a commonly expected pattern it might be a good idea to call it out in the PX4 docs on external modes?

[...]

PR #82, opened 2025-03-05
Make TrajectorySetpointType more flexible

azerupi:
Hey @bkueng I made an attempt at implementing what we discussed in #78 

This makes the TrajectorySetpointType more flexible by allowing the user to set position, velocity, acceleration, yaw, and yaw rate individually. For this we introduce a new helper struct TrajectorySetpoint with a builder pattern to set the different values.

I might have overdone it by making all fields individually settable through the builder pattern but I thought it would be more convenient than having to mix field access and builder pattern.

We tested this really quickly in simulation and it seems to work.

Let me know if you want me to change anything.

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/trajectory.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/experimental/trajectory.hpp
@@ -16,6 +16,8 @@ namespace px4_ros2
  *  @{
  */
 
+struct TrajectorySetpoint;
+
 /**
  * @brief Setpoint type for trajectory control
  *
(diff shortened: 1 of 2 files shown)


bkueng (commented):
This works, looks very nice, thanks.
There's some minor CI failures for doxygen (https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/13674589326/job/38476321502?pr=82#step:4:254) and formatting (https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/13674589318/job/38476321585?pr=82). The other failures are unrelated.

azerupi:
Thanks!
I believe I fixed both CI issues.

I'm running Jazzy and it looks like the version of `ament_uncrustify` I have installed formats code differently. Do you know how I can run the right version? For now I just made the changes by hand.

PR #84, opened 2025-03-18
Add vtol transition command

mahima-yoga:
- VTOL class through which user can directly call the `.to_multicopter()` or `to_fixedwing()` methods 
- Includes logic to check when the last command was sent before sending it again. 
- Logic is consistent with implementation in [px4 commander ](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/Commander.cpp#L355-L363)


mahima-yoga:
@GuillaumeLaine just tagged you in case there's anything that jumps out to you as an immediete no go. This PR is against another branch which will also be reviewed in its entirety later.

px4_ros2_cpp/src/control/vtol.cpp
```diff
@@ -0,0 +1,81 @@
...
+        _system_id = msg->system_id; 
+        _component_id = msg->component_id; 
+        _vehicle_type = msg->vehicle_type;
+
+    });
+
+ }
+ 
+ 
+ void VTOL::transition()
```
GuillaumeLaine, line 35: Let's check if the vehicle is capable of transitioning (i.e. is a vtol) before sending out the command, and warn the user otherwise

px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
```diff
@@ -0,0 +1,51 @@
...
+   void transition(); 
+   
+   private:
+   rclcpp::Node & _node;
+   rclcpp::Publisher<px4_msgs::msg::VehicleCommand>::SharedPtr _vehicle_command_pub;
+   rclcpp::Subscription<px4_msgs::msg::VehicleStatus>::SharedPtr _vehicle_status_sub; 
+
+   px4_msgs::msg::VehicleStatus::UniquePtr _vehicle_status_msg;
+
+   uint _system_id{1};
```
GuillaumeLaine, line 43: I'd avoid instantiating this to a possible sys/comp id. We may end up in a case where the user sends a transition command before the first vehicle status arrives, which would cause the wrong message payload to be sent out.

Instead the transition method should reject the user request if we don't have enough information to send out the command

px4_ros2_cpp/src/control/vtol.cpp
```diff
@@ -0,0 +1,81 @@
...
+        px4_msgs::msg::VehicleCommand cmd;
+        
+        cmd.command = px4_msgs::msg::VehicleCommand::VEHICLE_CMD_DO_VTOL_TRANSITION; 
+        cmd.param1 = px4_msgs::msg::VtolVehicleStatus::VEHICLE_VTOL_STATE_FW;
+
+        cmd.param2 = 0;
+   
+        cmd.source_system = _system_id; 
+        cmd.target_system = _system_id; 
+        cmd.source_component = _component_id;
```
GuillaumeLaine, line 49: Not sure but it doesn't seem correct that the src and dst compids are the same. Does PX4 reject the command if the source compid is left empty?
I guess the issue here is that the app using the SDK is generally not a MAVLink component

px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
```diff
@@ -0,0 +1,51 @@
...
+ class VTOL
+ {
+ public:
+ 
+   explicit VTOL(Context & context);
+ 
+   /**
+    * VTOL Transition
+    */
+
```
GuillaumeLaine, line 33: It would be nice for the VTOL interface to provide the user a way to query the current vehicle state i.e. FW or MC

px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
```diff
@@ -0,0 +1,51 @@
...
+   void transition();
```
GuillaumeLaine, line 34: I'm not familiar with how PX4 handles VehicleCommand messages, but there seems to exist an acknowledgement message [`VehicleCommandAck` ](https://github.com/PX4/PX4-Autopilot/blob/main/msg/versioned/VehicleCommandAck.msg) existing.
Ideally we should listen for that ack, and surface that reply to the user

px4_ros2_cpp/src/control/vtol.cpp
mahima-yoga: Interesting, I didn't know this was a MAVLink thing. Do you reckon it's fine to remove both source ids? That's what I have done for now -> still seems to work.

px4_ros2_cpp/src/control/vtol.cpp
mahima-yoga: addressed

px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
mahima-yoga: Unfortunately (as far as I have seen) the Ack message doesn't report anything for VTOL transitions

px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
mahima-yoga: addressed

px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
```diff
@@ -0,0 +1,67 @@
+/****************************************************************************
+ * Copyright (c) 2023 PX4 Development Team.
```
RomanBapst, line 2:
```suggestion
 * Copyright (c) 2025 PX4 Development Team.
```

PR #88, opened 2025-03-27
message_compatibility_check: add airspeed_validated

sfuhrer:
As we now have airspeed_validated versioned, in the [DDS bridge](https://github.com/PX4/PX4-Autopilot/pull/24302/files) and are using it in the SDK.

### DIFF
--- a/px4_ros2_cpp/CMakeLists.txt
+++ b/px4_ros2_cpp/CMakeLists.txt
@@ -46,6 +46,7 @@ set(HEADER_FILES
         include/px4_ros2/odometry/global_position.hpp
         include/px4_ros2/odometry/local_position.hpp
         include/px4_ros2/odometry/angular_velocity.hpp
+        include/px4_ros2/odometry/airspeed_validated.hpp
         include/px4_ros2/utils/frame_conversion.hpp
         include/px4_ros2/utils/geodesic.hpp
         include/px4_ros2/utils/geometry.hpp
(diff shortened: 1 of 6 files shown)


bkueng:
Can you also add a [telemetry class](https://github.com/Auterion/px4-ros2-interface-lib/tree/main/px4_ros2_cpp/include/px4_ros2/odometry) that makes it available?

sfuhrer:
@mahima-yoga could you help me out here and take it over?

mahima-yoga:
I see that we need a `RequirementFlags` object to check that the mode has the telemetry available, [like here](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/odometry/angular_velocity.cpp#L16-L20). 

To me, none of the existing flags seem relevant for this subscription, which are based on the `ArmingCheckReply.msg` [(msg definition)](https://github.com/PX4/px4_msgs/blob/main/msg/ArmingCheckReply.msg) . 

@sfuhrer do you know if there is an existing combination of mode flags required to have airspeed data? 

Otherwise, @bkueng can the flags be left empty?

sfuhrer:
> @sfuhrer do you know if there is an existing combination of mode flags required to have airspeed data?

Currently not, airspeed is, as far as the Commander is concerned, always optional.

bkueng (approved):

PR #96, opened 2025-04-08
added possibility to force avoiding local position checks

feoz0:
Added the possibility, when declaring the mode requirements in the mode constructor, to specify to avoid the local position checks. 

The need for this option is to implement a mode similar to altitude mode in PX4. For this scope we need to send local position setpoints on the Z axis and horizontal acceleration setpoints on X and Y (which practically map directly to angle setpoint in PX4). So this mode is angle-controlled in XY and position-controlled on Z. This mode should therefore work with just a valid altitude estimate and not the full local position validity. With this option the user can specify to avoid local position checks and force the drone to switch to the mode when only altitude is valid.

This is the solution I came up with but if you have idea on improvements that can be made or other ways to do the same please feel free to advise.

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/common/requirement_flags.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/common/requirement_flags.hpp
@@ -59,6 +59,7 @@ struct RequirementFlags
   bool home_position{false};
   bool prevent_arming{false}; ///< If set, arming is prevented when the mode is selected
   bool manual_control{false};
+  bool local_position_is_optional{false};
 };
 
 } // namespace px4_ros2
--- a/px4_ros2_cpp/include/px4_ros2/common/requirement_flags.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/common/requirement_flags.hpp
@@ -59,7 +59,6 @@ struct RequirementFlags
   bool home_position{false};
   bool prevent_arming{false}; ///< If set, arming is prevented when the mode is selected
   bool manual_control{false};
-  bool local_position_is_optional{false};
 };
 
 } // namespace px4_ros2
--- a/px4_ros2_cpp/src/components/mode.cpp
+++ b/px4_ros2_cpp/src/components/mode.cpp
@@ -251,16 +251,11 @@ void ModeBase::updateModeRequirementsFromSetpoints()
     requirements.angular_velocity |= config.rates_enabled;
     requirements.attitude |= config.attitude_enabled;
     requirements.local_alt |= config.altitude_enabled;
-    requirements.local_position |= config.velocity_enabled;
-    requirements.local_position |= config.position_enabled;
     requirements.local_alt |= config.climb_rate_enabled;
 
     if (!config.local_position_is_optional) {
         requirements.local_position |= config.velocity_enabled;
         requirements.local_position |= config.position_enabled;
-    }
-    else {
-      requirements.local_position = false;
     }
   }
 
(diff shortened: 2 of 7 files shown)


bkueng:
I think I would rather pass it through the relevant constructors as optional flag:
- https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/odometry/local_position.cpp#L12
- https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/control/setpoint_types/experimental/trajectory.cpp#L12 and add something like `local_position_is_optional` to https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/common/setpoint_base.hpp#L45

feoz0:
Hi @bkueng! I changed the strategy to include` local_position_is_optional` as an optional boolean to the class constructor of trajectory setpoint and odometry local position. I will most probably have the CI issue another time, but other than that let me know if there is anything I could change

px4_ros2_cpp/src/control/setpoint_types/experimental/trajectory.cpp
```diff
@@ -90,8 +91,16 @@ SetpointBase::Configuration TrajectorySetpointType::getConfiguration()
...
-  config.position_enabled = true;
+
+  if(!_local_position_is_optional){
+    config.position_enabled = true;
+    config.velocity_enabled = true;
+  }
+  else {
+    config.position_enabled = false;
+    config.velocity_enabled = false;
+  }
```
bkueng, line 102: Did you test this? This means the velocity controller is disabled.
You'll need to add a flag `local_position_is_optional` to the [Configuration struct](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/include/px4_ros2/common/setpoint_base.hpp#L45). And then use it here: https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/components/mode.cpp#L247

px4_ros2_cpp/include/px4_ros2/common/requirement_flags.hpp
```diff
@@ -59,6 +59,7 @@ struct RequirementFlags
   bool home_position{false};
   bool prevent_arming{false}; ///< If set, arming is prevented when the mode is selected
   bool manual_control{false};
+  bool local_position_is_optional{false};
```
bkueng, line 62: This one should not be needed
```suggestion
```

px4_ros2_cpp/src/components/mode.cpp
```diff
@@ -254,6 +254,10 @@ void ModeBase::updateModeRequirementsFromSetpoints()
     requirements.local_position |= config.velocity_enabled;
     requirements.local_position |= config.position_enabled;
     requirements.local_alt |= config.climb_rate_enabled;
+
+    if (config.local_position_is_optional) {
+    requirements.local_position = false;
+    }
```
bkueng, line 260:
```suggestion
    if (!config.local_position_is_optional) {
        requirements.local_position |= config.velocity_enabled;
        requirements.local_position |= config.position_enabled;
    }
```

feoz0: about this, I commited what you proposed but add an else that still sets the requirement to false in case the local_position_is_optional is true, otherwise it won't cancel the requirement

bkueng: No, the `else`  should not be needed, as it will be `false` already by default. The logic here should only set flags to `true`, otherwise the behavior might change depending on the calling order, which will be surprising and unexpected behavior.

feoz0:
Hi @bkueng, I fixed the `else` issue, actually I didn't delete the previous part (you'll see in the commit) and that's what's causing the issue. Thanks a lot for your help

bkueng (approved):
Thanks for your efforts, this looks good now. I applied it in https://github.com/Auterion/px4-ros2-interface-lib/commit/bb5050ad1268a6a552ddb0c8caa1d1a22e11f8b1

PR #97, opened 2025-04-08
Added class definition to read sensor data 

feoz0:
For some specifics application reading the vehicle accelerometer/gyro data directly in FRD frame is very useful. The local position class in the library already allows to read acceleration data but it's data that is already filtered (instead, this class allows to read the data that  has been only scaled and offset-compensated) and in NED frame. This PR adds a SensorData class with two methods to retrieve the 3-dimensional vector of the accelerometer/gyro data in FRD frame.  

One of the use case for this PR is to detect the freefall of the drone using accelerometer data.


feoz0:
currently we can only detect freefall using the vehicle_local_position NED filtered acceleration data. From what we found using non-filtered data gives us more robustness on the fall detection because instead of looking when the detected acceleration is close to the gravity vector (NED case), we look for when the acceleration vector is close to zero.

bkueng:
I remembered for freefall there's also a flag already in VehicleLandDetected: https://github.com/PX4/PX4-Autopilot/blob/main/msg/versioned/VehicleLandDetected.msg#L5.
Did you check if that works for you?

feoz0:
Hi @bkueng, the method they use in that flag for detecting is way too simple, our algorithm has been tested much more and is more reliable. In addition to that PX4 freefall boolean is used just as a safety measure, it is not at all used in freefall detection for a throw launch (which instead looks at the horizontal speed)

bkueng:
Ok makes sense. For the sake of dependencies, I'd still prefer if you keep this class in your own code for now. Specifically, `SensorCombined` is currently an unversioned internal PX4 topic.
Or alternatively we could add it to a separate directory and make sure it's flagged appropriately.

PR #99, opened 2025-04-08
Added possibility to force arm the drone and skip preflight checks 

feoz0:
This PR includes the possibility to skip the preflight checks if forced from the executor (when the arm function is called). The use case of this is for implementing a turtle mode where the drone is able to flip itself when upside-down. In this case usual preflight checks will not allow arming (drone is upside down) but arming is needed to perform the flip. 

This is already implemeted correspondingly in PX4 but wasn't implemented on this library.

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
@@ -97,7 +97,7 @@ public:
   void land(const CompletedCallback & on_completed);
   void rtl(const CompletedCallback & on_completed);
 
-  void arm(const CompletedCallback & on_completed, const bool & run_preflight_checks = true);
+  void arm(const CompletedCallback & on_completed, bool run_preflight_checks = true);
   void waitReadyToArm(const CompletedCallback & on_completed);
   void waitUntilDisarmed(const CompletedCallback & on_completed);
 
--- a/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
@@ -97,7 +97,7 @@ public:
   void land(const CompletedCallback & on_completed);
   void rtl(const CompletedCallback & on_completed);
 
-  void arm(const CompletedCallback & on_completed);
+  void arm(const CompletedCallback & on_completed, const bool & run_preflight_checks = true);
   void waitReadyToArm(const CompletedCallback & on_completed);
   void waitUntilDisarmed(const CompletedCallback & on_completed);
 
--- a/px4_ros2_cpp/src/components/mode_executor.cpp
+++ b/px4_ros2_cpp/src/components/mode_executor.cpp
@@ -280,11 +280,9 @@ void ModeExecutorBase::arm(const CompletedCallback & on_completed, const bool & 
     px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
     1.f);
 
-  if (!run_preflight_checks)
-  {
-    result = sendCommandSync(px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
-    1.f, 21196.f);
-  }
+  const float param2 = run_preflight_checks ? NAN : 21196.f;
+  const result = sendCommandSync(px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
+    1.f, param2);
 
   if (result != Result::Success) {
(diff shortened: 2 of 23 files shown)


px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
```diff
@@ -97,7 +97,7 @@ class ModeExecutorBase
   void land(const CompletedCallback & on_completed);
   void rtl(const CompletedCallback & on_completed);
 
-  void arm(const CompletedCallback & on_completed);
+  void arm(const CompletedCallback & on_completed, const bool & run_preflight_checks = true);
```
bkueng, line 100:
```suggestion
  void arm(const CompletedCallback & on_completed, bool run_preflight_checks = true);
```

px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -269,17 +269,23 @@ void ModeExecutorBase::rtl(const CompletedCallback & on_completed)
...
     on_completed(Result::Success);
     return;
   }
 
-  const Result result = sendCommandSync(
+  Result result = sendCommandSync(
     px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
     1.f);
 
+  if (!run_preflight_checks)
```
bkueng, line 283:
```suggestion
  if (!run_preflight_checks && result != Result::Success)
```
You could also use this:
```
const float param2 = run_preflight_checks ? NAN : 21196.f;
const result = sendCommandSync(px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
    1.f, param2);
```

bkueng (commented):
I hope people won't abuse this, but it makes sense for your use-case.

feoz0:
Hi @bkueng thanks for you suggestions! I have updated the PR, let me know if there is any other improvement I can make!

px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
feoz0: I understand the concern, to me it's still something valuable for this kind of modes that require skipping every flight checks

px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -269,17 +269,21 @@ void ModeExecutorBase::rtl(const CompletedCallback & on_completed)
   scheduleMode(ModeBase::kModeIDRtl, on_completed);
 }
 
-void ModeExecutorBase::arm(const CompletedCallback & on_completed)
+void ModeExecutorBase::arm(const CompletedCallback & on_completed, const bool & run_preflight_checks)
```
bkueng, line 272:
```suggestion
void ModeExecutorBase::arm(const CompletedCallback & on_completed, bool run_preflight_checks)
```

px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -269,17 +269,21 @@ void ModeExecutorBase::rtl(const CompletedCallback & on_completed)
...
 {
   if (_is_armed) {
     on_completed(Result::Success);
     return;
   }
 
-  const Result result = sendCommandSync(
+  Result result = sendCommandSync(
     px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
     1.f);
```
bkueng, line 281: You can remove this one now
```suggestion
```

px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -434,7 +437,8 @@ ModeExecutorBase::ScheduledMode::ScheduledMode(
   const std::string & topic_namespace_prefix)
 {
   _mode_completed_sub = node.create_subscription<px4_msgs::msg::ModeCompleted>(
-    topic_namespace_prefix + "fmu/out/mode_completed" + px4_ros2::getMessageNameVersion<px4_msgs::msg::ModeCompleted>(), rclcpp::QoS(
+    topic_namespace_prefix + "fmu/out/mode_completed" +
+      px4_ros2::getMessageNameVersion<px4_msgs::msg::ModeCompleted>(), rclcpp::QoS(
```
bkueng, line 441: Now CI complains about all these changes. 
You're possibly using a different ament_uncrustify version from CI? CI is using ROS Humble.
Sorry for the troubles.

bkueng:
I fixed the remaining issues and applied it to `main`: https://github.com/Auterion/px4-ros2-interface-lib/commit/7f6dccca91a983c09772a6cd1a8666d371678fad
Thanks for the contribution.

PR #100, opened 2025-04-08
Added possibility to disarm the drone from a mode executor

feoz0:
This PR introduces the disarm function to the mode executor. it is also possible to force the disarm. This is highly linked to this [PR](https://github.com/Auterion/px4-ros2-interface-lib/pull/99) and the use case is the same

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
@@ -98,7 +98,7 @@ public:
   void rtl(const CompletedCallback & on_completed);
 
   void arm(const CompletedCallback & on_completed);
-  void disarm(const CompletedCallback & on_completed, const bool & forced = false);
+  void disarm(const CompletedCallback & on_completed, bool forced = false);
   void waitReadyToArm(const CompletedCallback & on_completed);
   void waitUntilDisarmed(const CompletedCallback & on_completed);
 
--- a/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
@@ -98,6 +98,7 @@ public:
   void rtl(const CompletedCallback & on_completed);
 
   void arm(const CompletedCallback & on_completed);
+  void disarm(const CompletedCallback & on_completed, const bool & forced = false);
   void waitReadyToArm(const CompletedCallback & on_completed);
   void waitUntilDisarmed(const CompletedCallback & on_completed);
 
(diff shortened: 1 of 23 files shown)


px4_ros2_cpp/include/px4_ros2/components/mode_executor.hpp
```diff
@@ -98,6 +98,7 @@ class ModeExecutorBase
   void rtl(const CompletedCallback & on_completed);
 
   void arm(const CompletedCallback & on_completed);
+  void disarm(const CompletedCallback & on_completed, const bool & forced = false);
```
bkueng, line 101:
```suggestion
  void disarm(const CompletedCallback & on_completed, bool forced = false);
```

px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -290,6 +290,36 @@ void ModeExecutorBase::arm(const CompletedCallback & on_completed)
...
+void ModeExecutorBase::disarm(const CompletedCallback & on_completed, const bool & forced)
+{
+  if (!_is_armed) {
+    on_completed(Result::Success);
+    return;
+  }
+
+  Result result;
+
+  if (forced)
```
bkueng, line 302: This would be a bit simpler:
```
const float param2 = forced ? 21196.f : NAN;
const result = sendCommandSync(px4_msgs::msg::VehicleCommand::VEHICLE_CMD_COMPONENT_ARM_DISARM,
    0.f, param2);

```

feoz0:
Thanks @bkueng for the suggestions, I have updated the PR. Let me know if there's any other change to make

px4_ros2_cpp/src/components/mode_executor.cpp
```diff
@@ -290,6 +290,28 @@ void ModeExecutorBase::arm(const CompletedCallback & on_completed)
     [this](const px4_msgs::msg::VehicleStatus::UniquePtr & msg) {return _is_armed;}, on_completed);
 }
 
+void ModeExecutorBase::disarm(const CompletedCallback & on_completed, const bool & forced)
```
bkueng, line 293:
```suggestion
void ModeExecutorBase::disarm(const CompletedCallback & on_completed, bool forced)
```

feoz0:
Hi @bkueng, I have a different ROS version I believe so the code style doesn't match. how can we fix this?

bkueng:
You could use docker, but I can also quickly do it to save you that effort.

bkueng (approved):
CI is using this: https://github.com/Auterion/px4-ros2-interface-lib/blob/main/.github/workflows/lint.yml#L20 on ubuntu 22.04.
Since this isn't the first time someone had issues with that, I'm considering switching to clang-format. That should be more stable.

PR #104, opened 2025-04-24
[against another PR] VTOL Transition acceleration setpoint computation 

mahima-yoga:
Add function to compute acceleration setpoint during a vtol transition.   `TrajectorySetpoint.msg` sets: 
- `acceleration = {0.f, 0.f, NAN}` during front transition and decelerates during back transition

### DIFF
--- a/px4_ros2_cpp/src/control/vtol.cpp
+++ b/px4_ros2_cpp/src/control/vtol.cpp
@@ -15,9 +15,9 @@ static constexpr float CONSTANTS_ONE_G = 9.80665f; // m/s^2
 namespace px4_ros2
 {
 
-VTOL::VTOL(Context & context)
-: _node(context.node())
-{  
+VTOL::VTOL(Context & context, const VTOLConfig & config)
+: _node(context.node()), _config(config)
+{
   _vehicle_command_pub = _node.create_publisher<px4_msgs::msg::VehicleCommand>(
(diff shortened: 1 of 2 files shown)


px4_ros2_cpp/src/control/vtol.cpp
```diff
@@ -118,5 +133,61 @@ void VTOL::to_fixedwing()
...
+float VTOL::compute_pitch_setpoint_during_backtransition(){
+
+  float deceleration_setpoint = VT_B_DEC_MSS; 
+
+  const Eigen::Vector2f velocity_xy_direction = {std::cos(_vehicle_heading), std::sin(_vehicle_heading)}; 
+  const Eigen::Vector2f back_transition_end_pos = _backtransition_commanded_position + BACK_TRANS_END_DIST * velocity_xy_direction; 
+  const float dist_to_end_pos_in_moving_direction = back_transition_end_pos.dot(velocity_xy_direction); 
+
+  if (dist_to_end_pos_in_moving_direction > __FLT_EPSILON__) {
+    deceleration_setpoint = _vehicle_velocity_xy.squaredNorm() / (2.f * dist_to_end_pos_in_moving_direction);
```
RomanBapst, line 164: @mahima-yoga I would not bother doing this, but would just use a default deceleration setpoint and add a method to update it.

RomanBapst (approved):

PR #105, opened 2025-04-30
OdometryAirspeed: remove airspeed_sensor_measurment_valid.

mahima-yoga:
see commit.

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/odometry/airspeed.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/odometry/airspeed.hpp
@@ -50,15 +50,6 @@ public:
     const px4_msgs::msg::AirspeedValidated & aspd = last();
     return aspd.true_airspeed_m_s;
   }
-
-  bool airspeedSensorMeasurementValid() const
-  {
-    if (!lastValid()) {
-      return false;
-    }
-    const px4_msgs::msg::AirspeedValidated & aspd = last();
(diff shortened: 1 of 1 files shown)


sfuhrer (approved):
How is it done for other odometry messages? Do they normally have this "MeasurementValid()" method or was this a one-off for airspeed?
If it's a general thing then removing it is worse and we should think twice about it, if not then it's a no-brainer.

mahima-yoga:
>How is it done for other odometry messages? Do they normally have this "MeasurementValid()" method or was this a one-off for airspeed?

@sfuhrer seems to be pretty much a 1:1 to the message definition. So if the px4 message has some `is_valid` field, it's extracted (if relevant). 

So, for example, the attitude class doesn't have it. Local position does, because it has the `xy_valid` `z_valid` `v_xy_valid` `v_z_valid` fields.

sfuhrer:
> > How is it done for other odometry messages? Do they normally have this "MeasurementValid()" method or was this a one-off for airspeed?
> 
> @sfuhrer seems to be pretty much a 1:1 to the message definition. So if the px4 message has some `is_valid` field, it's extracted (if relevant).
> 
> So, for example, the attitude class doesn't have it. Local position does, because it has the `xy_valid` `z_valid` `v_xy_valid` `v_z_valid` fields.

Okay that's enough, so we don't need this method necessarily.

PR #106, opened 2025-05-07
Remove deprecated ament_target_dependencies

jmackay2:
ROS 2 is deprecating `ament_target_dependencies` with a preference to switch to `target_link_libraries`. This updates the CMakeLists to use `target_link_libraries`. Everything builds correctly.


bkueng:
Thanks, there's a CI failure for ROS rolling that looks real: https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/14872738628/job/41778729582?pr=106
Here it was still passing (last week): https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/14755037966/job/41421058571

bkueng (approved):
Thanks for checking.

PR #107, opened 2025-05-08
Replace apt with apt-get to reduce warnings in the CI logs

jmackay2:
Right now the CI has the following warning messages:
```
 WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

This removes some of the warning messages by using `apt-get` instead of `apt` to make the CI logs cleaner to read.


bkueng (approved):
Thanks, I always thought `apt-get` was deprecated in favour of `apt`, but apparently it's still the preferred tool in scripts.

PR #109, opened 2025-05-16
Add message versioning to subscription class

mahima-yoga:
Added message version suffix to ROS2 topic name in subscription class. See commit.


GuillaumeLaine:
@mahima-yoga Currently the topic string passed to the `Subscription` constructor assumes that the topic name is fully qualified, including the version suffix. For example see the [`OdometryAirspeed` construction](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/odometry/airspeed.cpp#L13-L15). As I see it, this change would cause the version to be appended twice to the topic name.

mahima-yoga:
@GuillaumeLaine oops, thanks! Closed the PR

PR #111, opened 2025-06-04
Switch to self-hosted runners

DanMesh:
Update the CI to use self-hosted runners

- [x] Requires https://github.com/Auterion/ros-debian-workflow/pull/11

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -12,7 +12,7 @@ defaults:
 jobs:
   build_and_test:
     name: Build and test
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


GuillaumeLaine:
@DanMesh shall we rebase this? New px4_msgs should fix CI, and we can get this in (:

DanMesh:
Let's see what happens...

GuillaumeLaine (approved):
Not sure why doxygen is suddenly failing on that bit that was not changed.
Regardless this should fix it: https://github.com/Auterion/px4-ros2-interface-lib/pull/138

This looks good to merge in the meantime thanks!

PR #114, opened 2025-06-05
add jazzy to ci

bastianhjaeger:


bkueng (approved):
Can you remove foxy?

bastianhjaeger:
> Can you remove foxy?

Could you, please? I am afk today.

bastianhjaeger:
> Can you remove foxy?

Correction: Didn't we say we might still need foxy for Payload Manager for a some time.

@GuillaumeLaine as well?

bkueng:
Ok we can do it later. Newest AOS uses Ubuntu 22.04 as base with ROS Humble.

PR #118, opened 2025-06-10
add jazzy to ci & change runners

bastianhjaeger:


bkueng:
@bastianhjaeger can we get this in?

PR #122, opened 2025-06-17
Add rmw_zenoh_cpp support by using RIHS01 hashes

PetervdPerk-NXP:
When using PX4 with rmw_zenoh_cpp with PX4/PX4-Autopilot#24476
We use RIHS01 hashes instead to check type support

This patch fixes that so that px4-ros2-interface-lib works with rmw_zenoh_cpp

### DIFF
--- a/px4_ros2_cpp/CMakeLists.txt
+++ b/px4_ros2_cpp/CMakeLists.txt
@@ -20,6 +20,41 @@ find_package(rclcpp REQUIRED)
 find_package(px4_msgs REQUIRED)
 find_package(eigen3_cmake_module REQUIRED)
 find_package(Eigen3 REQUIRED)
+find_package(ament_index_cpp REQUIRED)
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
+
+if($ENV{ROS_DISTRO} STREQUAL "humble")
+    add_compile_definitions(ROS2_HUMBLE)
+endif()
+
+ament_index_get_resource(MESSAGE_FILES "rosidl_interfaces" px4_msgs)
+
+# Split the message files into a list
+string(REPLACE "\n" ";" msg_file_list "${MESSAGE_FILES}")
+
+set(PX4_MSG_FILES "")
+foreach(file ${msg_file_list})
+  if("${file}" MATCHES "^msg/(.*\\.msg)$")
+    string(REGEX REPLACE "^msg/" "" cleaned_file "${file}")
+    list(APPEND PX4_MSG_FILES "${cleaned_file}")
+  endif()
+endforeach()
+
+# Convert list to Python-style list string
+string(REPLACE ";" "','" MESSAGE_NAMES_QUOTED "${PX4_MSG_FILES}")
+set(MESSAGE_NAMES_PYTHON "[\'${MESSAGE_NAMES_QUOTED}\']")
+
+add_custom_command(
+  OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/px4_all_messages.hpp
+  COMMAND ${Python3_EXECUTABLE} -m em
+    -D message_names=\"${MESSAGE_NAMES_PYTHON}\"
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/components/px4_all_messages.hpp.em
+    > ${CMAKE_CURRENT_BINARY_DIR}/px4_all_messages.hpp
+  DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/src/components/px4_all_messages.hpp.em
+  COMMENT "Generating px4_all_messages.hpp using empy"
+)
+
+add_custom_target(generate_px4_header DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/px4_all_messages.hpp)
 
 include_directories(include SYSTEM ${Eigen3_INCLUDE_DIRS})
 set(HEADER_FILES
@@ -116,6 +151,8 @@ add_library(px4_ros2_cpp
         src/utils/geodesic.cpp
         src/utils/map_projection_impl.cpp
 )
+add_dependencies(px4_ros2_cpp generate_px4_header)
+target_include_directories(px4_ros2_cpp PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
 target_link_libraries(px4_ros2_cpp PUBLIC ament_index_cpp::ament_index_cpp Eigen3::Eigen rclcpp::rclcpp ${px4_msgs_TARGETS})
 
 ament_export_targets(px4_ros2_cpp HAS_LIBRARY_TARGET)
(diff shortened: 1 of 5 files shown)


px4_ros2_cpp/src/components/message_compatibility_check.cpp
```diff
@@ -229,89 +350,116 @@ bool messageCompatibilityCheck(
...
-        if (response.success) {
-          if (response.message_hash != expected_message_hash) {
-            mismatched_topics += "\n  - " + message_to_check.topic_name;
+
+  if(px4_ros2::is_rmw_zenoh()) {
+    for (const auto & entry : all_px4_ros2_messages) {
+      auto publishers_info = node.get_publishers_info_by_topic(entry.topic_name);
+      const rosidl_type_hash_t *pub_hash;
+
+      if (!publishers_info.empty()) {
```
bkueng, line 362: So if not advertised yet this will silently succeed (?)

px4_ros2_cpp/src/components/message_compatibility_check.cpp
```diff
@@ -229,89 +350,116 @@ bool messageCompatibilityCheck(
...
-      case RequestMessageFormatReturn::ProtocolVersionMismatch:
-        // Error already reported
-        return false;
-      case RequestMessageFormatReturn::GotReply:
-        if (response.success) {
-          if (response.message_hash != expected_message_hash) {
-            mismatched_topics += "\n  - " + message_to_check.topic_name;
+
+  if(px4_ros2::is_rmw_zenoh()) {
+    for (const auto & entry : all_px4_ros2_messages) {
```
bkueng, line 358: This does not respect `messages_to_check`. Can we change the interface to include the type support?
Then we can also remove https://github.com/Auterion/px4-ros2-interface-lib/blob/5cf3e71d61d5fcced0983a422ac2ff8386674d32/px4_ros2_cpp/include/px4_ros2/components/message_compatibility_check.hpp#L12

px4_ros2_cpp/src/components/message_compatibility_check.cpp
PetervdPerk-NXP: Well if there's no publisher advertising it, there's no way to check if the hashes match.
Unless there should also be a check if all these topics are being published?

px4_ros2_cpp/src/components/message_compatibility_check.cpp
PetervdPerk-NXP: I could loop over both I suppose
```
for (const auto & entry : all_px4_ros2_messages) {
    bool found = false;
    for (const auto & message_to_check : messages_to_check) {
        if(entry.topic_name == message_to_check.topic_name) {
            found = true;
            // Do RIHS01 check
        }
    }
    if(!found) {
        // Not sure what to put here because original code hash mismatched_topics.
        // But it's not mismatched just simply not found.
    }
}
```

bkueng: I was thinking something like this: `bool messageCompatibilityCheck(rclcpp::Node & node, const std::vector<TopicTypeSupport> & messages_to_check=all_px4_ros2_messages`

px4_ros2_cpp/src/components/message_compatibility_check.cpp
bkueng: I don't have a better solution, just that this does not provide any guarantee. The minimum we can do is document it in the API docs.

px4_ros2_cpp/src/components/message_compatibility_check.cpp
bkueng: I can do the testing

PetervdPerk-NXP:
@bkueng I've updated the checking mechanism for Zenoh, removing the need of a fixed list to check.

Instead now I'll pull all topics and check the compatibility using the msgs type name instead of topic name.
Since for zenoh we don't have the tight coupling with topic names anymore.

Regarding the `ALL_PX4_ROS2_MESSAGES` discussion we had before I suppose this is only applicable for the UXRCE route, for Zenoh I just generate now the `px4_all_messages.hpp` that has all types and I can check them on a running system.

px4_ros2_cpp/CMakeLists.txt
```diff
@@ -20,6 +20,42 @@ find_package(rclcpp REQUIRED)
...
+  endif()
+endif()
+
+# Convert list to Python-style list string
+string(REPLACE ";" "','" MESSAGE_NAMES_QUOTED "${PX4_MSG_FILES}")
+set(MESSAGE_NAMES_PYTHON "[\'${MESSAGE_NAMES_QUOTED}\']")
+
+add_custom_command(
+  OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/px4_all_messages.hpp
+  COMMAND /usr/bin/empy
```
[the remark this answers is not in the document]
PetervdPerk-NXP: Good catch using `${Python3_EXECUTABLE} -m em` now instead.

bkueng:
Looks good, there's some formatting CI issues: https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/18158344509/job/51767970938?pr=122
You can[ install pre-commit](https://github.com/Auterion/px4-ros2-interface-lib?tab=readme-ov-file#development) to have that checked/fixed locally.

PetervdPerk-NXP:
> Looks good, there's some formatting CI issues: https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/18158344509/job/51767970938?pr=122 You can[ install pre-commit](https://github.com/Auterion/px4-ros2-interface-lib?tab=readme-ov-file#development) to have that checked/fixed locally.

Somehow my local `pre-commit` behaves differently then the CI, I just did the fixes manually. Could you start the CI again to check if it's okay now?

bkueng:
Linting passes now, there's still a build error though:
```
   [ 18%] Building CXX object CMakeFiles/px4_ros2_cpp.dir/src/components/message_compatibility_check.cpp.o
  In file included from /__w/px4-ros2-interface-lib/px4-ros2-interface-lib/ros_ws/src/5ksdtmjmzqb/px4-ros2-interface-lib/px4_ros2_cpp/src/components/message_compatibility_check.cpp:18:
  /__w/px4-ros2-interface-lib/px4-ros2-interface-lib/ros_ws/build/px4_ros2_cpp/px4_all_messages.hpp:5:10: fatal error: px4_msgs/msg/.hpp: No such file or directory
      5 | #include <px4_msgs/msg/.hpp>
        |          ^~~~~~~~~~~~~~~~~~~
  compilation terminated.
```

bkueng:
Let's make it easier, I gave you write permissions.

PetervdPerk-NXP:
I've got the CI happy now, Unfortunately it seems that ROS2 Humble doesn't support type hash, thus no type checking will occur when using Zenoh on ROS2 Humble. Therefore It's recommended when using Zenoh to use with ROS2 Jazzy or later

PR #123, opened 2025-06-19
Add change motor spin direction function

benchaso:
This function is useful for turtle mode after a crash. In fact, when the drone has flipped (tilt angle > 90°), there is a need to turtle mode. To do that, there is a need to turn some motors in the opposite direction to the default one for a normal flight.


bkueng:
This is a bit specific to be added to `ModeExecutorBase`. I'd like to avoid bloating that class.
Can you add something like a separate generic `CommandHandler` class and then a `ChangeMotorSpinDirectionCommand` class that uses it?

PR #133, opened 2025-07-14
Extend default action for takeoff with a takeoff altitude

eyeam3:
The takeoff mode in PX4 ascends to a default height above the takeoff point.

With this change, an optional altitude argument is added to enable takeoffs to a mission specific altitude (MSL).

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/mission/actions/default_actions.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/mission/actions/default_actions.hpp
@@ -60,7 +60,11 @@ public:
     const std::shared_ptr<ActionHandler> & handler, const ActionArguments & arguments,
     const std::function<void()> & on_completed) override
   {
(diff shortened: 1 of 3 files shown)


px4_ros2_cpp/include/px4_ros2/mission/actions/default_actions.hpp
```diff
@@ -60,7 +60,11 @@ class Takeoff : public ActionInterface
     const std::shared_ptr<ActionHandler> & handler, const ActionArguments & arguments,
     const std::function<void()> & on_completed) override
   {
-    handler->runMode(ModeBase::kModeIDTakeoff, on_completed);
+    if (arguments.contains("altitude")) {
+      handler->runModeTakeoff(arguments.at<float>("altitude"), NAN, on_completed);
+    } else {
+      handler->runMode(ModeBase::kModeIDTakeoff, on_completed);
```
bkueng, line 66: For consistency I'd also use `runModeTakeoff` here.

PR #136, opened 2025-08-04
[DRAFT] feat(mission_executor): expose flight mode info in action handler

alessandroferraioli95:
We need a way from an action to set some extra information about the current action, so those can gets exposed externally if wanted.

This commit adds an interface from the ActionHandler to set/clear the flight mode info into the mission_execturo


px4_ros2_cpp/include/px4_ros2/mission/mission_executor.hpp
```diff
@@ -425,6 +444,35 @@ class ActionHandler
...
+      RCLCPP_WARN(_mission_executor._node.get_logger(), "ActionHandler is not valid anymore");
+      return;
+    }
+    _mission_executor.setActivityInfo(activityInfo);
+  }
+
+  /**
+   * @brief Clears the activity information.
+   *        This method sets the activity info of the mission executor to `std::nullopt`. 
+   *        It should be called after an action is completed to ensure the state is properly reset.
```
bkueng, line 466: Let's call this automatically when switching to the next waypoint, or when the mission is deactivated or aborted.

bkueng (approved):

PR #153, opened 2025-10-01
Fix deprecated rclcpp::spin_some

jmackay2:
When building on rolling I get the following errors in multiple locations:
```
error: ‘void rclcpp::spin_some(Node::SharedPtr)’ is deprecated: use SingleThreadedExecutor::spin_some instead [-Werror=deprecated-declarations]
   62 |       rclcpp::spin_some(_node);
      |       ~~~~~~~~~~~~~~~~~^~~~~~~
```

This change updates the `spin_some` to use the `SingleThreadedExecutor::spin_some` instead as suggested in the deprecation warning.

### DIFF
--- a/px4_ros2_cpp/test/integration/global_navigation.cpp
+++ b/px4_ros2_cpp/test/integration/global_navigation.cpp
@@ -31,7 +31,7 @@ protected:
   void SetUp() override
   {
     _node = initNode();
-    
+
     _executor.add_node(_node);
 
     _global_navigation_interface = std::make_shared<GlobalPositionMeasurementInterface>(*_node);


px4_ros2_cpp/test/integration/global_navigation.cpp
```diff
@@ -122,7 +125,7 @@ TEST_F(GlobalPositionInterfaceTest, fuseAll) {
       break;
     }
 
-    rclcpp::spin_some(_node);
+    _executor.spin_some();
```
bkueng, line 128: Can you generally use `spin_node_some` to preserve the existing behavior? The difference is that it will also remove the node again.


```suggestion
    _executor.spin_node_some(_node);
```

bkueng (commented):
Thanks, looks good. Can you check the CI style failure? https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/18150108683/job/51659148541?pr=153

If you wanted to preserve the exact same behavior as before, you could use `executor.spin_node_some(_node);`.

bkueng (approved):
> This has been updated. It looks like many of the build failures are now related to px4_msgs::msg::ArmingCheckRequest not having a member named valid_registrations_mask even though it does. This seems unrelated to the changes in this PR.

Correct, the package build CI still uses an older message version set.

> Good point. I don't have a preference, but I can change it if desired.

Let's add it like this, and if there's any problems we'll change it.

PR #154, opened 2025-10-03
Add speed weight to FW control configuration

Jaeyoung-Lim:
**Problem Description**
For fixed-wing vehicles, it is useful to set an airspeed setpoint with the height rate uncontrolled.

While the message exists, the ROS2 interface does not expose the speed weight.

**Solution**
This commit adds speed weight to the FWControlConfiguration

- Depends on: https://github.com/PX4/PX4-Autopilot/pull/25709


Jaeyoung-Lim:
@mahima-yoga Done

px4_ros2_cpp/include/px4_ros2/control/setpoint_types/fixedwing/lateral_longitudinal.hpp
```diff
@@ -204,6 +206,15 @@ struct FwControlConfiguration
     target_climb_rate = target_climb_rate_sp;
     return *this;
   }
+
+  FwControlConfiguration & withSpeedWeight(float speed_weight_sp)
```
bkueng, line 210: Can you add some API docs? Just seeing this it would not be clear to me what it does, and what values to use.

PR #156, opened 2025-10-07
Expose throttle direct for FwLateralLongitudinal Setpoint

Jaeyoung-Lim:
**Problem Description**
When conducting fixed-wing vehicle performance estimation, one would measure the sink rate of the vehicle at different airspeed setpoints and throttle settings. However, the `FWLateralLongitudinalSetpoint` does not expose the throttle_direct even though the underlying setpoint message has a field for throttle_direct.

**Solution**
This PR exposes throttle_direct for the FWLateralLongitudinalSetpoint. 

Tested in SITL. Sending an airspeed setpoint and fixed throttle results in a stable sinkrate.
- Log: https://review.px4.io/plot_app?log=f0dac6f3-7f8d-49a2-860b-cb2c282820f5

**Additional Context**
- This PR is implemented on top of https://github.com/Auterion/px4-ros2-interface-lib/pull/154

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/fixedwing/lateral_longitudinal.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/setpoint_types/fixedwing/lateral_longitudinal.hpp
@@ -119,6 +119,7 @@ struct FwLateralLongitudinalSetpoint
   std::optional<float> altitude_msl;
   std::optional<float> height_rate;
   std::optional<float> equivalent_airspeed;
+  std::optional<float> throttle_direct;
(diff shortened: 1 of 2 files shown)


mahima-yoga:
The reason throttle (and pitch) weren’t included in the initial implementation is that they were considered too “low level” for this particular setpoint. I don’t remember exactly, but I believe even the PX4 fixed-wing controller rework only added them to ensure takeoff still worked correctly (or something along those lines).

We decided that if you want to control these directly, you can use the attitude setpoint instead. What’s the argument against using the attitude setpoint directly in this case?

PR #171, opened 2025-12-10
Add collision avoidance mode for LiDAR-based obstacle detection

satyampd2025-droid:
Features:
- Subscribes to drone_path_planner waypoints
- Integrates with PX4 external mode system
- Yaw control to face direction of travel
- Health monitoring integration

### DIFF
--- a/examples/cpp/modes/collision_avoidance/include/mode.hpp
+++ b/examples/cpp/modes/collision_avoidance/include/mode.hpp
@@ -1,0 +1,150 @@
+#pragma once
+
+#include <px4_ros2/components/mode.hpp>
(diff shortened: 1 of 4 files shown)


examples/cpp/modes/collision_avoidance/include/mode.hpp
```diff
@@ -0,0 +1,150 @@
...
+  {
+    // TODO: Re-enable health check after testing
+    // For now, always consider healthy if we receive status messages
+    (void)msg;
+    _planner_healthy = true;
+  }
+};
+
+// Alias for backward compatibility
+using CollisionAvoidanceMode = WaypointFollowerMode;
```
bkueng: Why is this needed?

PR #176, opened 2026-01-10
Add odometry subscription helper

rocketSzw:
Provide Odometry wrapper exposing NED position/velocity and attitude fields, set required PX4 data, and register the new header/source in the build.


bkueng (approved):
Nice, thanks

PR #186, opened 2026-03-15
fix(vtol): make status staleness timeout configurable (default 5s)

PavelGuzenfeld:
## Summary

- Replace hardcoded 2s `VtolVehicleStatus` staleness timeout with configurable `status_timeout` in `VTOLConfig` (default: 5s)
- Fix uninitialized `_last_vtol_vehicle_status_received` clock type causing exception when calling `toMulticopter()`/`toFixedwing()` before receiving any status message
- Add unit tests for `VTOLConfig` and `VTOL` transition behavior

## Problem

**Timeout too tight:** PX4's `vtol_att_control` publishes `VtolVehicleStatus` at a minimum rate of 1 Hz (only faster on state changes). The hardcoded 2s staleness window tolerates only one missed message before silently refusing transition commands. Under real-world conditions (SITL CPU contention, DDS jitter, Gazebo rendering load), this causes intermittent transition failures.

**Clock mismatch crash:** `_last_vtol_vehicle_status_received` was default-constructed without a clock type, while `toMulticopter()`/`toFixedwing()` compare it against `_node.get_clock()->now()`. If called before any `VtolVehicleStatus` is received, this throws `"can't subtract times with different time sources"` instead of returning `false`.

## Changes

**`vtol.hpp`:**
- Add `std::chrono::seconds status_timeout{5}` to `VTOLConfig` with `withStatusTimeout()` builder
- Initialize `_last_vtol_vehicle_status_received` with `_node.get_clock()->get_clock_type()` (matching the existing pattern for `_last_pitch_integrator_update`)

**`vtol.cpp`:**
- Replace `< 2s` with `< _config.status_timeout` in both `toMulticopter()` and `toFixedwing()`

**`test/unit/vtol.cpp`:** (new)
- `VTOLConfig.DefaultValues` — verify all defaults including `status_timeout`
- `VTOLConfig.WithStatusTimeout` — builder pattern
- `VTOLConfig.BuilderChaining` — chaining multiple builders
- `VTOLTest.InitialStateUndefined` — initial state is `Undefined`
- `VTOLTest.TransitionFailsWithoutStatus` — returns `false` before any status received (previously crashed)
- `VTOLTest.TransitionSucceedsWithFreshStatus` — works after receiving status
- `VTOLTest.ToFixedwingSucceedsFromMulticopter` — MC→FW transition
- `VTOLTest.StateTransitions` — all 4 state transitions via published messages
- `VTOLTest.CustomTimeoutIsUsed` — verifies configurable timeout works (1s timeout, wait 1.1s, transition fails)

## Why 5s default

- 1 Hz publication rate → messages every ~1s
- 5s allows up to 4 missed heartbeats before considering state unknown
- Still detects genuinely dead connections within a reasonable window
- Configurable for tighter or looser bounds via `VTOLConfig{}.withStatusTimeout(...)`

## Test results

```
[==========] 76 tests from 14 test suites ran. (7009 ms total)
[  PASSED  ] 76 tests.
Summary: 77 tests, 0 errors, 0 failures, 0 skipped
```

Fixes #185

### DIFF
--- a/px4_ros2_cpp/CMakeLists.txt
+++ b/px4_ros2_cpp/CMakeLists.txt
@@ -225,6 +225,7 @@ if(BUILD_TESTING)
             test/unit/mission_execution.cpp
             test/unit/modes.cpp
             test/unit/shared_subscription.cpp
+            test/unit/vtol.cpp
             test/unit/utils/frame_conversion.cpp
             test/unit/utils/geodesic.cpp
             test/unit/utils/geometry.cpp
--- a/px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/control/vtol.hpp
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <chrono>
+
 #include <Eigen/Core>
 #include <px4_msgs/msg/vehicle_command.hpp>
 #include <px4_msgs/msg/vehicle_local_position.hpp>
@@ -19,6 +21,8 @@ struct VTOLConfig {
   float back_transition_deceleration_setpoint_to_pitch_i{
       0.1f}; /**< Backtransition deceleration setpoint to pitch I gain  [rad s/m]. */
   float deceleration_integrator_limit{0.3f};
+  std::chrono::seconds status_timeout{
+      5}; /**< Maximum age of VtolVehicleStatus before state is considered unknown [s]. */
 
   VTOLConfig& withBackTransitionDeceleration(const float back_transition_deceleration)
   {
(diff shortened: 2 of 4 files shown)


bkueng (commented):
Makes sense, thanks. Could you check the minor linting failure? https://github.com/Auterion/px4-ros2-interface-lib/actions/runs/23118171272/job/67202181290?pr=186

bkueng (approved):
Looks good. You can install pre-commit to automatically fix the linting: https://github.com/Auterion/px4-ros2-interface-lib?tab=readme-ov-file#development

bkueng (approved):

Issue #108, opened 2025-05-13
Support sending GPS_GLOBAL_ORIGIN and DO_SET_HOME commands


tedlin0913:
Hello!
Thank you for this great work. My goal is to use this library together with VIO for position estimation in order to fly my drone in GNSS-denied areas. To achieve this, I followed the px4 guide on [enable auto mode with local position](https://docs.px4.io/main/en/ros/external_position_estimation.html#enabling-auto-modes-with-a-local-position), which suggests sending the `VEHICLE_CMD_SET_GPS_GLOBAL_ORIGIN` command. In addition, `VEHICLE_CMD_DO_SET_HOME` can be used to set the home position once the global position estimate is available.
I am now trying to figure out how to send these two VehicleCommands correctly using this library. So far, I have attempted to send the commands via `sendCommandSync` in my mode executor `ModeExecutorTest`, which inherits from `ModeExecutorBase`. The functions I written for sending the command looks like this:

```c++
void ModeExecutorTest::setHome(const CompletedCallback &on_completed)
{
    if (is_home_set_)
    {
        on_completed(Result::Success);
        return;
    }
    // TODO: check if global position is available before sending the command? 
    const Result result = sendCommandSync(
        px4_msgs::msg::VehicleCommand::VEHICLE_CMD_DO_SET_HOME,
        1.f);

    // TODO: maybe subscribe HomePosition.msg to check if home position is availible?

    if (result != Result::Success)
    {
        on_completed(result);
        return;
    }
}

void ModeExecutorTest::setGlobalOrigin(const CompletedCallback &on_completed)
{
    if (is_origin_set_)
    {
        on_completed(Result::Success);
        return;
    }
   
    // VEHICLE_CMD_SET_GPS_GLOBAL_ORIGIN does not send back a VehicleCommandAck.
    sendCommandSync(
        px4_msgs::msg::VehicleCommand::VEHICLE_CMD_SET_GPS_GLOBAL_ORIGIN,
        NAN, NAN, NAN, NAN,
        0.0, 0.0, 0.0);
}
```
VEHICLE_CMD_DO_SET_HOME sends back a VehicleCommandAck, whereas VEHICLE_CMD_SET_GPS_GLOBAL_ORIGIN does not.

I plan to use these commands in `runState`. Two follow up questions occur: 
1. How can I check, inside the mode executor, when the global position estimate becomes available, before setting the home position?
[...]

PR #189, opened 2026-03-18
feat(vehicle_state): add HomePositionSetter utility

PavelGuzenfeld:
## Summary
Add a `HomePositionSetter` utility class for sending `VEHICLE_CMD_DO_SET_HOME` and `VEHICLE_CMD_SET_GPS_GLOBAL_ORIGIN` commands to PX4, as suggested by @bkueng in #108.

- `setHomeToCurrentPosition()` — sets home to current vehicle position
- `setHome(lat, lon, alt)` — sets home to a specific global coordinate
- `setGpsGlobalOrigin(lat, lon, alt)` — sets the EKF reference point

All methods send the command and **wait for an ACK** (3 retries × 300 ms timeout), returning a `Result` enum.

The ACK-waiting logic is extracted into a reusable `VehicleCommandSender` utility (`px4_ros2/utils/vehicle_command_sender.hpp`), modeled after the pattern in `ModeExecutorBase::sendCommandSync()`. This helper accepts a configurable command topic, so `ModeExecutorBase` can also be migrated to it in a follow-up.

Ref #108

## Test plan
- [x] `colcon build --packages-select px4_ros2_cpp` passes (ROS2 Jazzy)
- [x] Unit tests: `VehicleCommandSender` (accepted / rejected / timeout) + `HomePositionSetter` (all 3 methods) with fake ACK responder
- [x] Integration test: verifies all 3 methods against PX4 SITL
- [x] SITL integration test (PX4 main + Gazebo + XRCE-DDS agent):
  - `setGpsGlobalOrigin(47.397742, 8.545594, 488.0)` — ACK cmd=100000 result=0
  - `setHomeToCurrentPosition()` — ACK cmd=179 result=0, home=47.397971,8.546164
  - `setHome(47.398, 8.546, 490)` — ACK cmd=179 result=0, home=47.398000,8.546000,490.00

### DIFF
--- a/px4_ros2_cpp/CMakeLists.txt
+++ b/px4_ros2_cpp/CMakeLists.txt
@@ -207,6 +207,7 @@ if(BUILD_TESTING)
             test/integration/mode.cpp
             test/integration/mission.cpp
             test/integration/mode_executor.cpp
+            test/integration/home_position_setter.cpp
             test/integration/overrides.cpp
     )
     if(TARGET integration_tests)
@@ -229,6 +230,7 @@ if(BUILD_TESTING)
             test/unit/mission_execution.cpp
             test/unit/modes.cpp
             test/unit/shared_subscription.cpp
+            test/unit/vehicle_command_sender.cpp
             test/unit/utils/frame_conversion.cpp
             test/unit/utils/geodesic.cpp
             test/unit/utils/geometry.cpp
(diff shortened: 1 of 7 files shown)


bkueng (commented):
Nice, looks good.

We should probably also wait and check for an ack (this isn't important for `PeripheralActuatorControls`, as this is a topic that's sent regularly).
As it is done here: https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/components/mode_executor.cpp#L121 (that could be moved into a common helper).

bkueng (approved):
Looks good, will you do the follow-up as well (replacing ModeExecutorBase::sendCommandSync())?

PR #199, opened 2026-05-07
feat(mission): add setSkipMessageCompatibilityCheck to MissionExecutor

presidentplant-snu:
Add `setSkipMessageCompatibilityCheck` to `MissionExecutor` , which propagates to the underlying `MissionModeExecutor` so callers can skip the FMU message compatibility check before `doRegister()`. 

Previously couldn't be skipped since all the relevant methods and/or members were either `protected` or `private`.


bkueng (commented):
Looks good. Can you check CI?

presidentplant-snu:
I have fixed the clang-format issue. My bad.

bkueng (approved):
Thanks

PR #202, opened 2026-05-18
fix(lateral_longitudinal): add boolean in class constructors to set local position requirement

mahima-yoga:
### Problem 
As with the trajectory setpoint type, the fixed wing lat/long setpoint exposes inputs at different levels in the controller. Some of these inputs don't require a valid local position - but the mode won't execute because the requirement is placed on the entire setpoint type. 

### Solution 
Adds a boolean in the class constructor to set the requirement on local position. This avoids local position being required when using lower level inputs in the setpoint type. 

Mimics what is already done for the [trajectory setpoint](https://github.com/Auterion/px4-ros2-interface-lib/blob/main/px4_ros2_cpp/src/control/setpoint_types/experimental/trajectory.cpp#L96), as the same problem exists there.


bresch (approved):
Having a boolean looks at bit ugly to me, but this is apparently done the same way elsewhere

PR #203, opened 2026-05-22
fix(wait_for_fmu): wait for matched subscriptions before returning

TSC21:
## Solved problem

The ROS 2 integration tests intermittently fail on PX4 CI with `ModesTest.runModeTests` and `HomePositionSetter` timing out after 120 s. The failures are non-deterministic and have been observed across multiple PR runs on `PX4/PX4-Autopilot` over the past several weeks.

Root cause is a DDS startup discovery race in the integration harness:

1. `waitForFMU()` returns the instant it receives one `vehicle_status` sample from PX4. This proves the FMU-out data writer for `vehicle_status` has matched our reader, but it says nothing about any *other* topic.
2. The test then immediately calls `_vehicle_state.sendCommand(...)` to request a mode change (`VEHICLE_CMD_SET_NAV_STATE`).
3. The publisher for `/fmu/in/vehicle_command_v<N>` may not yet have matched PX4's subscription. With BEST_EFFORT QoS the message is silently discarded -- there is no retry, no queueing, no error.
4. The mode change never takes effect, the gtest 120 s wall timer expires, the job fails.

Why `runMission` happens to pass: it waits for one extra `setOnVehicleStatusUpdate` callback before its first `sendCommand`, buying ~100 ms of extra discovery time.

Evidence from a failing ULG (PX4 PR [#27297](https://github.com/PX4/PX4-Autopilot/pull/27297) CI run):
260 `vehicle_status` samples were received over 129.7 s, but only one `vehicle_command` -- and that one was the startup `COM_RC_LOSS_T` sent by PX4 itself, never the test's mode-change.

## Solved by this PR

Block in the integration test's `VehicleState` constructor (and in the production `VehicleCommandSender::sendCommandSync`) until at least one matched subscription / matched publisher count is non-zero before returning, with a polling timeout (5 s for the integration helper, 3 s for the production sender; 50 ms intervals). Discovery normally completes within 100-500 ms in CI, so the new wait is essentially free on the happy path and prevents the race entirely.

Two files touched:

* `px4_ros2_cpp/test/integration/util.cpp` -- the actual cause of the flaky CI failures.
* `px4_ros2_cpp/src/utils/vehicle_command_sender.cpp` -- has the same pattern (BEST_EFFORT publisher followed by an immediate publish) and was already attempting to wait for one direction (ack publisher count) but not the other (command subscriber count). Symmetrise it.

## Validation

A focused rclpy reproducer was built to exercise the race in isolation (without needing PX4 SITL or Gazebo). It mimics the exact publish/ subscribe topology and QoS of the integration test:

* **FMU side**: publishes `vehicle_status` immediately, creates the `vehicle_command` subscriber after a configurable delay (500 ms - 1 s) to provoke the race.
* **Test side**: subscribes `vehicle_status`, creates `vehicle_command` publisher, on first `vehicle_status` callback either (a) publishes immediately (no fix) or (b) polls `get_subscription_count()` until matched (this fix).

Results over 10 fresh `ROS_DOMAIN_ID` iterations each:

| Condition                              | Drops | Mean elapsed |
|----------------------------------------|------:|-------------:|
| no fix, no FMU sub delay               | 0/10  | 1.93 s       |
| no fix, 500 ms FMU sub delay           |**10/10**| 1.71 s     |
| **with fix**, 500 ms FMU sub delay     | 0/10  | 2.02 s       |
| **with fix**, 1000 ms FMU sub delay    | 0/10  | 2.75 s       |

The "no fix + 500 ms delay" case reproduces the CI failure shape exactly (test publishes its command, FMU never receives it). With the fix applied, the race is eliminated even at 2x the realistic delay.

## Test plan

- [x] `colcon build --packages-select px4_ros2_cpp` on ROS 2 Humble
- [x] Standalone DDS reproducer, 10 iter, 500 ms delay: 0/10 drops
- [x] Standalone DDS reproducer, 10 iter, 1 s delay: 0/10 drops
- [x] `clang-format -i` applied to both files
- [ ] Upstream CI (this PR)
- [ ] PX4 CI re-run on `PX4/PX4-Autopilot#27297` once this is merged

## Related

Soft dependency of [PX4/PX4-Autopilot#27297](https://github.com/PX4/PX4-Autopilot/pull/27297) -- the Windows-native SITL PR has been blocked on ROS Integration Test flakiness, which this fix resolves.

### DIFF
--- a/px4_ros2_cpp/src/utils/vehicle_command_sender.cpp
+++ b/px4_ros2_cpp/src/utils/vehicle_command_sender.cpp
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: BSD-3-Clause
  ****************************************************************************/
 
+#include <chrono>
 #include <px4_ros2/utils/vehicle_command_sender.hpp>
+#include <thread>
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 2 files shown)


bkueng (approved):
Looks good, thanks.

We could use `VehicleCommandSender` in the integration test as well.

veluneos:
Hi — I reproduced the before/after behavior around #203 and ran a structural comparison across repeated Reference/Target executions.

I used the pre-change commit as the Reference and the merged #203 change as the Target, with alternating run order (R1 → T1 → R2 → T2) to reduce run-order effects.

Across all four Reference→Target comparisons, the same runtime difference was surfaced consistently, with roughly 380–402 ms of unmatched coverage on the relevant response stream. The same-condition controls were substantially quieter (about 20.5 ms for Reference→Reference and 0.36 ms for Target→Target).

I’m testing VeluneOS Compare for this type of before/after runtime review. I’m not treating the result as an independent root-cause or correctness conclusion — the value I’m testing is whether the comparison can consistently localize where execution behavior diverges.

If useful, I can share the exact commits, reproduction setup, run matrix, and comparison outputs.

And if you have a current internal PX4/ROS 2 before/after change where this kind of review would be useful, I’d be interested in running the same comparison on one Reference/Target pair.

Issue #206, opened 2026-06-09
Rates setpoints fail to apply correctly with new PX4 Autopilot


maticpokorn:
## The problem

I tried to run a MPC rate controller mode on the latest PX4 and the rates setpoints the MPC published were not being applied correctly. The same mode worked perfectly fine on older PX4 versions. It seems to me that rate setpoints and other setpoint types were being applied interchangeably.

## How to reproduce

1. PX4 SITL on **`main`** (≥ `0eb14d64d5`)
2. Register an external mode with `RatesSetpointType` via `ModeBase::doRegister()`
3. Arm and select the mode
4. Publish rates setpoints

**Expected:** vehicle follows desired trajectory; only rate control active.

**Actual:** doesn't follow trajectory / erratic thrust; internal controllers fight external setpoints.

**Verify in PX4 shell:**

```bash
listener vehicle_control_mode
```

On broken setup you see `flag_control_attitude_enabled: true` and/or `flag_control_position_enabled: true` when you expect rates-only.

```bash
listener vehicle_rates_setpoint
```

Values jump between your MPC output and internal attitude-controller output.

## Suggested fix (library side)

On every mode **activation**, republish `config_control_setpoints` for the active setpoint type. On **deactivation**, mark setpoint types inactive so the next activation republishes cleanly.

Minimal change in `ModeBase::callOnActivate()` / `callOnDeactivate()`.

We implemented this locally:

- Fork: [TheLukaDragar/px4-ros2-interface-lib](https://github.com/TheLukaDragar/px4-ros2-interface-lib)
- Commit: [`d8edcf1`](https://github.com/TheLukaDragar/px4-ros2-interface-lib/commit/d8edcf1) — `fix(mode): republish config_control_setpoints on mode activation`

bkueng:
This broke in https://github.com/PX4/PX4-Autopilot/pull/27372, and apparently was not tested well.

Your change makes sense, can you open a PR? The `republishActiveSetpointConfiguration` should not be needed, as the setpoint will immediately be activated in https://github.com/TheLukaDragar/px4-ros2-interface-lib/commit/d8edcf1#diff-7c6ab1a8341022f925ea2bc5b298c0dc2d577d7851e8e01890a96e208ca35c57L122-L124.

maticpokorn:
[...]

PR #207, opened 2026-06-15
Fix setpoint activation sequence (issue #206)

maticpokorn:
# Pull request: fix setpoint activation sequence

This is the fix to the issue described in #206.

## Problem description
Newer PX4 versions refuse control modes that predate mode activation. 

The problem was that `_setpoint_types[0]._active` was being set to `true` prematurely in `onRegistered`. This caused the `SetpointBase._should_activate_cb` method to be skipped, failing to publish a renewed `VehicleControlMode`.

## Overview of changes
Below is a quick overview of the lifecycle process before and after the changes proposed in this pull request:

| Lifecycle Step | Current Behavior | This PR |
| :--- | :--- | :--- |
| **`ModeBase.onRegistered`** | `VehicleControlMode` published.<br>`_setpoint_types[0]._active` is set to `true`. | `VehicleControlMode` published.<br>`_setpoint_types[0]._active` remains `false`. |
| **`ModeBase.callOnActivate`** | `VehicleControlMode` **not** published.<br>`_setpoint_types[0]._active` remains `true`. | `VehicleControlMode` published.<br>`_setpoint_types[0]._active` is set to `true`. |
| **`SetpointBase.onUpdate`** | `_active` is already `true`, so `should_activate_cb()` is **skipped**. | `_active` is now `true`, so `should_activate_cb()` is **skipped**. |

In the current version, no new `VehicleControlMode` gets published after `onRegistered`. This fix would work even if we omitted the changes in `callOnActivate`; in that case, `should_activate_cb` would simply be triggered and publish the `VehicleControlMode` during the very first update loop.

### DIFF
--- a/px4_ros2_cpp/include/px4_ros2/components/mode.hpp
+++ b/px4_ros2_cpp/include/px4_ros2/components/mode.hpp
@@ -206,6 +206,7 @@ class ModeBase : public Context {
   void updateModeRequirementsFromSetpoints();
   void setSetpointUpdateRateFromSetpointTypes();
   void activateSetpointType(SetpointBase& setpoint);
+  void deactivateAllSetpointTypes();
 
   std::shared_ptr<Registration> _registration;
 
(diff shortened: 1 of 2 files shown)


bkueng (approved):
Looks good, thanks

PR #210, opened 2026-07-16
feat: add reportEventExt() to send GCS events without failing the arming check

mahima-yoga:
Adds `reportEventExt()` to `HealthAndArmingCheckReporter`  (identical to `armingCheckFailureExt() `but leaving `can_arm_and_run` untouched). 

This lets external modes send informational or warning events to the ground station without preventing mode selection or forcing a fallback.


bkueng (commented):
There's a small CI error, otherwise looks good.

mahima-yoga:
>There's a small CI error, otherwise looks good.

I opened a seperate PR to fix the CI failure as it seems unrelated: https://github.com/Auterion/px4-ros2-interface-lib/pull/213

PR #211, opened 2026-07-16
chore: adopt a full ROS 2 build-farm release cycle

TSC21:
## What this is, and why it is needed

This makes `px4-ros2-interface-lib` installable and releasable as a normal, versioned ROS 2 package, on a predictable cadence, the same way `px4_msgs` already is. Today the library has no proper release process, so a downstream user cannot simply install a versioned build tied to a known PX4 version; they build from source or pull from an ad-hoc channel. This PR sets up the whole pipeline: version numbers aligned to the PX4 release line, a cross-distribution CI build-and-test matrix, Debian packaging, and automated releases that publish into the public ROS package index (`ros/rosdistro`) when a new PX4 message version ships. It changes no library source or behavior; it is packaging, CI and release infrastructure only.

Who it is for: anyone who wants to depend on this library as a released, versioned ROS 2 package rather than a source checkout, and the maintainers who need a repeatable, hands-off release process. It mirrors what [PX4/px4_msgs#68](https://github.com/PX4/px4_msgs/pull/68) adopted, adapted for a two-package source library (`px4_ros2_cpp` C++ and `px4_ros2_py` pybind11 bindings).

This work is contributed by [RIIS, LLC](https://www.riis.com).
RIIS_Black@2x

## Versioning

Versions now track the **PX4 Autopilot release line** (this PR sets `1.17.0`), matching what `px4_msgs` already does, so `px4_ros2_cpp`, `px4_ros2_py`, `px4_msgs` and PX4 all move on one axis and the compatibility story is simply "px4_ros2 1.17.x works with PX4 1.17 and px4_msgs 1.17.x". The in-tree `package.xml` version becomes authoritative (bloom reads it), replacing the placeholder-plus-tag-injection scheme.

## What this adds

**Packaging.** Both manifests carry the release-line version, `url` tags and a RIIS author credit; `px4_ros2_cpp` declares the previously-undeclared `python3-empy` build dependency that backs its `px4_all_messages.hpp` codegen; rosdep rule files cover Humble, Jazzy, Kilted, Rolling and Lyrical and map the off-farm `px4_msgs` and `px4_ros2_cpp` keys to their Debian names for the packaging container.

[...]

PR #212, opened 2026-07-16
feat: ROS 2-native flight SDK (RFC + px4_ros2_sdk P0 scaffold)

TSC21:
## What this is, and why it is needed

This adds a task-level, ROS 2-native flight SDK on top of PX4's existing low-level ROS 2 library (`px4_ros2_cpp`), so a companion application can command a PX4 vehicle through standard ROS 2 actions, services and topics instead of hand-wiring the low-level pieces itself. The first working step here is arm and takeoff; setpoint streaming, telemetry and richer maneuvers follow in later phases per the design.

Why it is needed: `px4_ros2_cpp` today is a low-level construction kit. Getting a vehicle to do something as simple as arm and take off means authoring mode and executor classes, binding setpoint types, wiring subscriptions and driving the command sender by hand. There is no ready-made, task-level surface a ROS 2 developer can pick up in a few minutes, so every team rebuilds the same plumbing and the safety, timing and frame-sensitive parts are easy to get subtly wrong. This SDK adds that missing layer, and it treats `px4_ros2_cpp` as the trusted core underneath, the same way `ros2_control` and Aerostack2 sit on top of their hardware layers. It does not bypass PX4 safety, and it adds no new messages across the bridge to the flight controller (the SDK's actions and services live entirely on the companion side).

Who it is for: the large pool of ROS 2 application developers, the people already building on Nav2, MoveIt 2 and behavior trees, who want to command a PX4 vehicle without first learning PX4 internals.

This PR has two parts:

- A design document (`doc/API_EXPANSION_RFC.md`) that lays out the full SDK and shows how each piece maps onto established ROS 2 patterns.
- A first working scaffold (`px4_ros2_sdk`) that proves the pattern end to end: a managed flight server exposing an arm service and a takeoff action, built on `px4_ros2_cpp`.

It is kept intentionally separate from the release-cycle PR #211. This work is contributed by [RIIS, LLC](https://www.riis.com).

RIIS_Black@2x

## The design (RFC)

`doc/API_EXPANSION_RFC.md` proposes the full `px4_ros2_sdk`: a managed lifecycle flight server, long-running maneuvers as action servers (with feedback, cancel and preempt), configuration and one-shot commands as services, setpoint streaming and telemetry as QoS-tuned topics modeled as `ros2_control`-style typed command and state interfaces, pluginlib behaviors and estimators, standard diagnostics, a behavior-tree-friendly surface, and an ergonomic `SimpleFlight` (C++) and `Drone` (Python) commander over that same graph. The patterns are drawn from the established ROS-native frameworks (Aerostack2, Nav2, MoveIt 2, ros2_control, Autoware, BehaviorTree.CPP, ROS 2 lifecycle and composition) rather than any single one.

The design is checked against the actual firmware rather than assumed: which topics the uXRCE-DDS bridge really exposes (so parameters, geofence and mission upload are shown as currently blocked rather than faked), a `setpoint_config` mismatch between firmware and library that scopes what the first phase can stream, and PX4's registered-external-mode liveness and fallback model. Section 4.5 works out how modes register (one hidden `OffboardSession` carries setpoint streaming; custom behaviors each own a registered mode within PX4's external-mode slot limit; takeoff, land and RTL register nothing and instead schedule PX4's own modes through `ModeExecutorBase`) and how failsafe requirements are handled per setpoint type, so a mode streaming rate commands does not inherit the position-loss failsafe of a mode it never uses.

## What is in the scaffold (P0)

- `px4_ros2_sdk_interfaces`: the `SetArmed` service and `Takeoff` action definitions.
- `FlightServer`, a managed lifecycle node: `on_configure` discovers the FMU and fails closed if it is absent; `on_activate` starts the `~/arm` service and the `~/takeoff` action, both backed by `px4_ros2_cpp`'s command sender; it is registered as a composable component (`flight_server_node`). The takeoff command leaves its unused command parameters as NaN so PX4's Navigator takes off at the current position rather than a literal (0, 0), and the action's worker thread is joined on deactivate and cleanup so it never outlives the command sender.
- `SimpleFlight`: a thin client-side commander (`arm`, `disarm`, `takeoff`) over that service and action, opening no bridge endpoint of its own.
- `sdk::Result`: reports the vehicle's acknowledgement of a command, not physical completion. In this phase the takeoff action completes on that acknowledgement; completion based on PX4's `mode_completed` rides the `ModeExecutorBase` path in a later phase.

The streaming setpoint interfaces, additional maneuver actions, the telemetry hub, pluginlib behaviors and the Python commander are proposed in the RFC and land in later phases.

## How it was validated

Built end to end in a `ros:jazzy-ros-base` container: `px4_msgs`, `px4_ros2_cpp`, `px4_ros2_sdk_interfaces` and `px4_ros2_sdk` all build, the `flight_server_node` component builds, and the `px4_ros2_sdk` unit tests pass (5/5). `clang-format`, `clang-tidy` and `catkin_pkg` all pass.

## Relationship to the release-cycle PR

This is intentionally separate from #211. Once #211's cross-distro `build.yml` lands, the `px4_ros2_sdk` packages should be added to that workflow's build and test set so they are exercised on the full matrix.

## Type of change

- [x] New feature (experimental package)
- [x] Documentation / design (RFC)

## Checklist

- [x] Layered additively on the core: no `px4_ros2_cpp` source or behavior changes, and no new `/fmu` endpoints beyond `px4_ros2_cpp`.
- [x] Builds and unit tests (5/5) pass in a container; `clang-format`, `clang-tidy` and `catkin_pkg` pass; SPDX headers on all sources.
- [x] Commits are signed off (DCO) and follow Conventional Commits.


TSC21:
@bkueng @mrpollo FYI and for review. Thank you!

bkueng:
From what I understand the main difference here is to provide an interface based on topics/services/actions and trying to use standard ROS2 messages as opposed to the existing C++/Python library interface (and please correct me if I'm wrong).

There are pros and cons to each. Specifically for the message-approach:
Pros:
- it is more ROS2-native, better integration with tooling/libraries like behavior trees
- potentially simpler to use
- higher degree of separation between components

Cons:
- adds another layer and more complexity
- more maintenance effort, more effort to change something
- potential loss of semantics and details, especially when using standardized ROS messages, e.g. for setpoints. This is a very important point to me as I've seen many times that people want to control very specific aspects
- with a lot of services/actions it can become harder to use as well, as certain services might only be usable under certain conditions or have dependencies

And then there are things that can be addressed independently (either in a library or message approach):
- NED vs FRD: the library uses explicit naming for clarity. However, we can go further and I'm fine with moving to FRD completely for the whole API. Or a middle ground would be to add typed classes with automatic type conversions.
- timesync can also be added either way
- diagnostics: how is it relevant?
- pluginlib: same, I don't see how that's even relevant?

Specifically one thing that I don't see being addressed here is mode registration: I see `OffboardSession` mentioned without any explanation. This is an important point that needs to be worked out.
Is the idea to support only one mode? If multiple are supported, how does registration work and how does it map to topics like the takeoff?
And how are setpoint types and telemetry handled? The current lib sets mode requirement flags based on what a mode uses. So for example a mode using rate setpoints will not failsafe if position is lost, whereas a mode using position setpoints would failsafe in that case.

> on top that adds no new DDS traffic

That part I don't understand, the example code already adds a new Takeoff and arming action/service?

Can you provide a bit more details on what you have in mind and how it should look like eventually? Like MAVROS?
What are specific use-cases?
And finally who is the targeted user? How does someone decide to use the library vs the messages?
And for setpoints specifically, is the idea to use the PX4 messages or also create a mapping?

And more general, for future, could you please take a critical pass yourself and edit AI generated output before submitting?

TSC21:
@bkueng thanks for the detailed pass! Going through it block by block.

> From what I understand the main difference here is to provide an interface based on topics/services/actions and trying to use standard ROS2 messages as opposed to the existing C++/Python library interface (and please correct me if I'm wrong).

Almost: it's not meant as an alternative to the library, it sits on top of it. `px4_ros2_cpp` stays the only thing talking to the FMU, and the SDK is a task-level ROS graph over it, so both surfaces coexist and the library keeps being the power-user path. Standard messages only where nothing is lost, which in practice means telemetry.

> potential loss of semantics and details, especially when using standardized ROS messages, e.g. for setpoints. This is a very important point to me as I've seen many times that people want to control very specific aspects

Fully agree, and the design already sides with you there: the streaming topics map 1:1 onto the `SetpointBase` subtypes (trajectory, attitude, rates, thrust/torque, plus the fixed-wing and rover ones), so nothing gets squeezed through a `Twist`. A `cmd_vel` lane exists as a convenience next to them, not instead of them. Same answer to your last question below: PX4 setpoint semantics preserved, and no lossy mapping.

> diagnostics: how is it relevant?

It's the standard surface a deployment uses to watch stream health and failsafe state when there's no GCS attached: `diagnostic_updater` FrequencyStatus on the republished streams plus the failsafe flags mapped into `/diagnostics`. Relevant for the robot integrator, not for the lib itself, and easy to defer if it feels like too much for a first cut.

> pluginlib: same, I don't see how that's even relevant?

It only matters if the packaged server should let users drop in their own behaviors by config instead of forking the node, the Nav2 pattern. If behaviors stay plain library-user code it disappears from the design. I don't hold it tightly...

> Specifically one thing that I don't see being addressed here is mode registration: I see `OffboardSession` mentioned without any explanation. This is an important point that needs to be worked out.

Agreed it needs its own section, RFC 4.1 only sketches it. The model so far: streaming rides one hidden registered mode (that's `OffboardSession`, registered `user_selectable` with internal-mode fallback intact), behaviors can each own a registered mode within the external-mode slot limit, and takeoff/land/rtl don't register anything, they schedule the PX4-native modes through `ModeExecutorBase` so completion rides `mode_completed`. The P0 code shortcuts that last part through a raw `VEHICLE_CMD_NAV_TAKEOFF` for now, which is exactly the kind of gap between RFC and scaffold I want to close.

> And how are setpoint types and telemetry handled? The current lib sets mode requirement flags based on what a mode uses. So for example a mode using rate setpoints will not failsafe if position is lost, whereas a mode using position setpoints would failsafe in that case.

That's the sharpest open problem in the whole thing and the RFC doesn't answer it today. One hidden mode multiplexing several setpoint types can't just take the union of requirements, or rate control would inherit the position failsafe. It likely ends up as per-type requirement configuration on the session (re-register or reconfigure on type switch), but I want to work that through properly in the RFC before anything streams.

> That part I don't understand, the example code already adds a new Takeoff and arming action/service?

You're right, that wording is wrong and it's on me. What I meant is no new traffic over the uXRCE bridge: no `fmu/in`/`fmu/out` endpoints beyond what the core lib already opens. The companion-side graph of course adds DDS entities (an action server alone is three services plus two topics). I'll reword the PR text, README and RFC to the invariant that's actually meant, "no new `/fmu` endpoints beyond `px4_ros2_cpp`".

> Can you provide a bit more details on what you have in mind and how it should look like eventually? Like MAVROS? What are specific use-cases? And finally who is the targeted user? How does someone decide to use the library vs the messages?

Not MAVROS, and I say that with years of MAVROS scars: no wholesale protocol mirror, but instead just a small curated task surface, closer to Nav2's simple commander sitting on its action graph. The target user is the companion-app author who today hand-rolls raw `px4_msgs` publishers to arm, take off and stream a setpoint, and gets QoS, frames or ack-vs-completion subtly wrong; you can think BT-driven inspection, quick Python mission scripting, fleet tooling. The rule of thumb is: writing a flight mode or feeding the estimator, use the library; writing an application that commands the vehicle and observes state, use the graph. This would probably be much closer to what MAVSDK does with MAVLink, but more ingrained in the ROS middleware and tooling.

> And more general, for future, could you please take a critical pass yourself and edit AI generated output before submitting?

To be straight about how this was made: I designed and programmed it together with an AI, and I did review the output before opening the PR; it wasn't submitted blind. Given the pushback I'm redoing that pass top to bottom with a much more critical eye, and it's already paying off: the P0 takeoff zero-fills the unused `VehicleCommand` params where `ModeExecutorBase::takeoff` uses NaN, and Navigator reads finite param5/param6 as a takeoff target lat/lon, so that gets fixed here along with the DDS wording. Is there something specific that read as unedited to you beyond what you've raised, or something else you'd want addressed first?

bkueng:
Ok I have a clearer picture now, thanks. I first got the impression the end result would be to expose the full library functionality. Something simple and more limited on top makes sense though. We'll just need to document it clearly.

At the same time if it ends up that you have to fill in raw PX4 setpoint messages yourself, and know when and how fast to publish them is a step back and an API (C++ or python) is IMO easier to use. Or maybe I miss how you envision that being used.

> It's the standard surface a deployment uses to watch stream health and failsafe state when there's no GCS attached: diagnostic_updater FrequencyStatus on the republished streams plus the failsafe flags mapped into /diagnostics. Relevant for the robot integrator, not for the lib itself, and easy to defer if it feels like too much for a first cut.

If you want to map the failsafe flags it's relevant. Although something that can be added independently later on.

> It only matters if the packaged server should let users drop in their own behaviors by config instead of forking the node, the Nav2 pattern. If behaviors stay plain library-user code it disappears from the design.

My understanding is that the server provides generic services/actions and if that is too limited, the library should be used instead. It would keep things simple.

> The model so far: streaming rides one hidden registered mode (that's OffboardSession, registered user_selectable with internal-mode fallback intact), behaviors can each own a registered mode within the external-mode slot limit, and takeoff/land/rtl don't register anything, they schedule the PX4-native modes through ModeExecutorBase so completion rides mode_completed

This should be worked out in detail then (or maybe you already have a clear picture). How to define, register and activate behaviors? And should these really be separate modes, as the number is limited on FMU side?
Mode activation generally requires an active mode executor, described [here](https://docs.px4.io/main/en/ros2/px4_ros2_control_interface#mode-executor). Of course it does not have to adhere to that, but I think from a predictability and safety perspective it's worth to do that. There's a big difference to previous offboard in that regard.

> Is there something specific that read as unedited to you beyond what you've raised, or something else you'd want addressed first?

It looked like a wall of AI text, with various sentences that do not make sense to me. Like:

> and the reachability wall drawn by dds_topics.yaml is surfaced rather than faked.

> not evasions, and each is confirmed absent rather than assumed absent.

> The firmware-versus-library skew that the first revision got wrong

PR #216, opened 2026-08-08
Add watchdog timeout notification callback

### LINKED ISSUE #195: Health checks crash companion computer on timeout
https://github.com/Auterion/px4-ros2-interface-lib/blob/365dd8807869fd81813de0415ec99e85ea021d59/px4_ros2_cpp/src/components/health_and_arming_checks.cpp#L89-L93

Because this is in a watchdog timer, there is no way to catch this exception on timeout. Desired behavior is just to prevent the mode from activating until this message is received.

Daniiiil1:
## Summary

- add an optional watchdog timeout callback to `HealthAndArmingChecks`
- expose the callback to derived mode implementations through `ModeBase`
- cancel the watchdog before notifying the application, allowing it to schedule mode re-creation without shutting down the ROS context
- retain the existing shutdown-and-throw behavior when no callback is configured

Fixes #195.

## Testing

Ran the repository pre-commit suite for all changed files, including mixed-line-ending checks, trailing-whitespace checks, and `clang-format`.


bkueng (approved):
Out of interest, do you know why you get timeouts in the first place? High CPU load?

PR #222, opened 2026-09-08
fix(registration): isolate reply subscription from executors

mrpollo:
## Summary

Allow mode registration while the node executor is already spinning.

## Problem

The executor and registration wait set can both claim the reply subscription, causing Jazzy to throw instead of registering the mode.

## Solution

Keep registration replies in a retained callback group excluded from automatic executor assignment. Cover registration and re-registration with both single-threaded and multi-threaded executors.


mrpollo:
Reproduced both new executor regressions against the original code: both throw the wait-set ownership exception. With the fix, 40 repeated regression invocations, all 85 unit tests, and all 13 SIH integration cases pass on Jazzy. No existing assertions, timeouts, or test filters were changed. This unblocks the fresh-source ROS integration migration in PX4/PX4-Autopilot#28596.

bkueng:
A waitset is also used in other places (e.g. `ModeBase::checkSetpointCompatibilityAndRequirements`). I assume these need to be updated too?

Did you also happen to test with ROS humble?

And after the PX4 side goes in we can also update https://github.com/Auterion/px4-ros2-interface-lib/blob/main/.github/workflows/px4_build.yml.

mrpollo:
@bkueng tested with Humble, Jazzy and Galactic, all working, I also took care of the remaining items I had originally dismissed, thanks that was a good call. Once this is in, and my work here https://github.com/PX4/PX4-Autopilot/pull/28596 is done I'll come back to this repo to update CI

bkueng (approved):
Nice, thanks @mrpollo!

